PR #16, opened 2026-07-31
hpsdr: add ppm frequency correction

reynico:
My Radioberry is slightly off frequency, and I haven't found a way to fix it through sdroxide

image

btw, thanks for the software. it is really awesome


dividebysandwich:
I was gonna ask if that's really a frequency offset issue - but then I googled and that's indeed a common problem! Good patch, thank you!

PR #26, opened 2026-08-02
session: persist volume, RX gain, drive and mic gain across restarts

reynico:
i found that some settings do not persist across restarts

### DIFF
--- a/crates/sdroxide-config/src/lib.rs
+++ b/crates/sdroxide-config/src/lib.rs
@@ -258,10 +258,10 @@ impl Settings {
     }
 }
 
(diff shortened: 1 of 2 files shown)


dividebysandwich:
There's a few small issues: 
1. A completely fresh installation would now result in a dead mic because Session::default() uses drive 0.0 and mic_gain 0.0, can you use RadioState::default() instead?
2. agc mode isn't persisted
3. Could you please also persist *both* tune_drive and drive?

Then rebase and you're good to go!

reynico:
thanks for the review! rebased to 98e8c00a034cbef87504724e9ee5a9001d4621d8

PR #27, opened 2026-08-02
radio: record both sides of a QSO, fix stale RX replay on unkey

reynico:
I was trying to record both sides of the QSO (RX on the left channel, TX on the right channel) when I noticed that, for a very brief period, on unkey, I heard back part of my transmission as fallback audio. 

I did a screen recording of a QSO where the unkey bug shows clearly: https://www.youtube.com/watch?v=1lBMSVBbU8o

So this PR addresses both things; just let me know if you prefer two separate pull requests to address each thing separately.

Audio recording looks like the following:

image

### DIFF
--- a/crates/sdroxide-radio/src/engine.rs
+++ b/crates/sdroxide-radio/src/engine.rs
@@ -32,7 +32,7 @@ use sdroxide_types::{
     ScanKind, ScanResume, SpectrumConfig, SpectrumFrame, TciServerConfig, TxMeters, Vfo,
 };
 
-use crate::recorder::Recorder;
+use crate::recorder::{Recorder, RecordingChannels};
 use crate::voice::VoiceKeyer;
 use crate::{Complex32, ControlUpdate, IqSource};
 
@@ -217,6 +217,12 @@ struct RxChain {
     out_buf_r: Vec<f32>,
     /// Resamples L and R together so the two can't drift a sample apart.
     stereo_rs: Option<StereoResampler>,
+    /// Post-squelch audio at `out_rate`, same shape as `out_buf`/`out_buf_r`,
+    /// but without the AF volume/mute scaling applied to those — the QSO
+    /// recorder taps this instead, so turning down the AF knob or hitting
+    /// mute doesn't touch the archived recording. See [`RxChain::take_rec_audio`].
+    rec_buf: Vec<f32>,
+    rec_buf_r: Vec<f32>,
 }
 
 impl RxChain {
@@ -246,6 +252,8 @@ impl RxChain {
             lr_out: Vec::new(),
             out_buf_r: Vec::new(),
             stereo_rs: None,
+            rec_buf: Vec::new(),
+            rec_buf_r: Vec::new(),
         };
         chain.build_for_mode(rx);
         chain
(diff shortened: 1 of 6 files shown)


crates/sdroxide-radio/src/engine.rs
```diff
@@ -517,6 +532,27 @@ impl StereoMixer {
             self.sub_q.drain(..cut);
         }
     }
+
+    /// Recording tap for TX audio: right channel, silence in the left.
+    /// Resampled from `TX_MONITOR_RATE` to match `rec_tap`.
+    fn push_tx(&mut self, tx: &[f32]) {
```
dividebysandwich, line 695: This needs the tuning signal to be included for IQ rigs. Otherwise pressing tune will offset the RX and TX audio, and every tune will make things worse

crates/sdroxide-radio/src/engine.rs
```diff
@@ -449,9 +449,14 @@ struct StereoMixer {
     main_q: Vec<f32>,
     sub_q: Vec<f32>,
     dropped: u64,
-    /// When recording, a copy of each interleaved L/R output frame is pushed
-    /// here.
+    /// When recording: RX in the left channel, TX audio in the right.
```
dividebysandwich: Can we make true mono possible for SWL/RX-only users? Also, could we optionally mix everything together into a mono file for users who don't want to listen to audio on one ear?

crates/sdroxide-radio/src/engine.rs
```diff
@@ -2772,7 +2770,9 @@ impl Engine {
         let path = dir.join(&name);
         match Recorder::start(path, self.audio_out_rate) {
             Ok((rec, prod)) => {
-                self.mixer.as_mut().expect("checked above").rec_tap = Some(prod);
+                let mixer = self.mixer.as_mut().expect("checked above");
```
dividebysandwich, line 3083: Hmm.. RX and TX recording behave differently. Right now RX is tapped post-volume, and TX is tapped post-mic gain with no AF volume applied. If I turn down the AF knob this will make the received half quiet, and muting would make it silent. Maybe we should do a pre-volume RX tap? What do you think?

crates/sdroxide-radio/src/engine.rs
```diff
@@ -478,15 +492,16 @@ impl StereoMixer {
...
-                    let _ = rec.push(l); // drop if the recorder stalls
-                    let _ = rec.push(r);
+            // Recording tap: RX left, sub receiver (if any) right, silence
+            // otherwise. Suppressed by rx_rec_enabled while transmitting.
+            if self.rx_rec_enabled {
+                if let Some(rec) = self.rec_tap.as_mut() {
+                    for i in 0..n {
+                        let l = self.main_q[i];
+                        let r = if dual { self.sub_q[i] } else { 0.0 };
+                        let _ = rec.push(l); // drop if the recorder stalls
```
dividebysandwich: Now that left and right are different, we'll need to guard against only one slot being free. Otherwise this can lead to left and right being swapped.

dividebysandwich (changes requested):

dividebysandwich:
Yes let's split this please. The unkey-fix is good, it just misses two backends: SmartSDR and SoapySDR. If you like we could do these in a separate PR too of course.

The recording change is a bit more of an open question for me. Your PR always records RX on the left. This changes things for everyone, and might not be what RX-only users want since they too now only get audio on the left channel (the comment in recorder.rs is wrong btw). 

There's some other comments too, see review

reynico:
sounds good, sorry for the noise! here's the unkey PR: https://github.com/dividebysandwich/sdroxide/pull/28

PR #41, opened 2026-08-08
modulator: follow rx filter for tx passband

reynico:
it seems like the TX passband bandwidth was around 2.7 kHz for SSB (technically, (150.0, 2850.0)). now one can change the TX filter following the RX filter but also by moving the passband during TX.


dividebysandwich:
Good catch, thanks for the PR!

PR #65, opened 2026-08-15
hpsdr: add configurable transmit-buffer cushion for lossy links

reynico:
I noticed some audio stutter/clipping when operating sdroxide over Wi-Fi, so a configurable buffer came in handy.


dividebysandwich:
There's a small discrepancy with the ring sizes which might bite some users in some edge cases, but again I'll merge and do the modification.

PR #89, opened 2026-08-17
Add an SWR guard: stop transmitting into a fault, and latch until acknowledged

FreespiritD:
sdroxide already reads the SWR the rig reports, shows it on the meter, announces it by speech and pushes it to TCI clients — but nothing ever acts on it. This adds the missing half: an optional guard that stops the transmission when the SWR is bad, and stays stopped until the operator says they have looked at it.

It is modelled on the equivalent feature in WSJT-X, which is where the request came from.

### What it does

While transmitting, the meter loop compares the rig's reported SWR against a limit. At or above it, the transmitter is unkeyed, transmit is latched out, and a persistent notice tells the operator to check the antenna. The latch clears only via `Command::ClearSwrTrip`, surfaced in the notice banner as an explicit button, so a fault cannot be dismissed into silence — if the antenna is still bad, the next over is stopped too.

`swr_guard` (on) and `swr_limit` (2.5:1) live in `config.toml` next to `tx_ham_only`, and both are editable in Settings → General.

**It is inert on any rig that does not report SWR**, so it costs nothing to leave enabled.

### Three decisions worth explaining

**Forward power is optional.** The first version required both an SWR and a forward-power reading before it would act. The CI-V driver sends `TxTelemetry { fwd_w: None, swr: Some(v) }`, so on every Icom that branch could never fire and the guard was dead code. It is now used only to reject a reading taken at essentially no power.

**There is no meaningful power floor.** The first version ignored readings below 5 W, to filter the key-up transient. That is self-defeating: a rig with SWR protection folds its power back *because* the SWR is bad — an IC-7300 dropped to 3 W into 7.5:1 — so any floor high enough to filter a transient is also high enough to gate out the emergency. Discarding the transient is a question of time, not power, so it is now done by time: SWR is ignored for the first six meter samples of each over (~0.2 s), and the trip needs six consecutive over-limit samples after that.

**The trip unkeys via `Command::SetPtt(false)`, not `deny_tx()`.** That helper only sets state flags, on its documented assumption that every deny happens before the transmitter keys — true of the rails in `sync_tx_state`, false here, where the rig is on the air. This follows the route the TCI starvation unkey already uses.

`TxState` gains `swr_guard`, `swr_limit` and `swr_tripped` so remote clients can show the setting and offer the acknowledgement, rather than each client reading a `config.toml` that may be on a different machine — the same reasoning that keeps the IARU region in the broadcast state. The engine clamps the limit to 1.1–10.0 so a client editing the field before the first state arrives cannot arm a threshold that refuses every transmission.

### Testing

`crates/sdroxide-radio/tests/swr_guard.rs`, six tests, all passing. They drive the real engine with a stand-in rig whose reported telemetry the test dictates, covering: the Icom shape (SWR present, forward power absent), a folded-back 3 W at 7.5:1, a good 1.3:1 left alone, a disarmed guard, a rig reporting no SWR, and a reading taken at genuinely 0 W. The first also asserts that the transmitter actually stops and that the next key-up is refused.

Also tested on air on an IC-7300 into a genuine 7.5:1: transmission stopped and the warning appeared.

One note in case it shows up in CI: `tests/scanner.rs::an_impossible_scan_is_refused_with_a_reason` fails for me, and it fails identically on an unmodified `main`, so it is not from this branch.

Happy to change any of the defaults, the wording, or where the settings live.

### DIFF
--- a/crates/sdroxide-config/src/lib.rs
+++ b/crates/sdroxide-config/src/lib.rs
@@ -88,6 +88,15 @@ pub struct Settings {
     pub server_port: u16,
     /// Refuse to transmit outside amateur bands.
     pub tx_ham_only: bool,
+    /// Abort the transmission and latch out further transmit when the rig
+    /// reports an SWR at or above [`Self::swr_limit`].
+    ///
+    /// Only rigs that actually measure SWR and report it over CAT/TCI can arm
+    /// this; on anything that leaves `TxTelemetry::swr` as `None` it is inert,
+    /// so it costs nothing to leave on.
+    pub swr_guard: bool,
+    /// The SWR ratio the guard trips at, e.g. `2.5` = 2.5:1.
+    pub swr_limit: f32,
     /// Preferred audio output device name; `None` = system default.
     pub audio_output: Option<String>,
     /// Preferred audio input (microphone) device name; `None` = system default.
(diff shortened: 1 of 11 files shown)


dividebysandwich:
This looks also good and is a great feature addition. I will probably make some minor modifications so that tuning doesn't trigger it too quickly, since tuning can last several seconds.

PR #90, opened 2026-08-17
Add a Test button for each logging service's credentials

FreespiritD:
Today the first sign that an eQSL password or a QRZ key is wrong is a QSO that failed to upload, hours after the contact. This adds a Test button beside each set of credentials in Settings, so the answer is available before anything depends on it.

Covers eQSL, QRZ Logbook, Club Log and LoTW.

### Nothing here writes

A credential check that inserted a dummy QSO to see whether the login worked would put a fictional contact in the operator's permanent log, and for the services that feed award programmes, somewhere it cannot easily be withdrawn from. Every branch uses the cheapest read endpoint the service publishes, and the upload functions are not reachable from `test_login`.

### The reply rules came from the services, not from the docs

This is the part I would want a reviewer to check hardest, so it is worth saying how it was arrived at. Every rule below is taken from what the service actually returned on 17 August 2026, verified for correct *and* wrong credentials wherever that could be done without risking a lockout.

- **eQSL** rejects with `Error: No such Username/Password found`, and reports an empty inbox as `You have no log entries`. My first version guessed at "invalid" / "not found" / "no qso" and was wrong in both directions at once: it would have passed a rejection and failed a good empty account. Verified both ways.
- **QRZ Logbook** has a `STATUS` action, which is the only endpoint of the four intended for exactly this. Verified against a real key and a bogus one (`invalid api key`).
- **Club Log**'s account is checked with `POST getadif.php` and `type=dxqsl`, so the whole log is not downloaded to answer the question.
- **LoTW** answers a bad login with its log-on page rather than an error, which is why the check looks for ADIF rather than for an error string. Worth knowing: LoTW logins are case-sensitive, and the username is the callsign in upper case, so a login that looks right can still be refused with nothing on the page saying why. That is precisely the kind of thing this button exists to surface.

### One gap, stated rather than hidden

**Club Log's API key is not checked, only the account.** The upload needs both. The documented key endpoint, `GET /dxcc?call=&api=&full=1`, returned `403 Forbidden` on every attempt here, with and without a User-Agent, seconds after the account call succeeded from the same host with the same key present in config. Rather than report a guess, the success message says `account accepted (API key not checked)`.

If there is a supported way to validate a key, I would happily add it and make that message unconditional.

### Smaller decisions

Pressing Test sends `SetNetworkConfig` first. Without that, a freshly pasted key would be checked against the previously applied one and reported wrong when it was merely unsaved, which is exactly when someone presses Test.

Results are not cached between sessions: a green tick from an hour ago says nothing about the password typed since.

The result is not forwarded to remote clients and the buttons are hidden there, matching how `RadioEvent::Notice` is handled, since the answer names the account the service recognised.

### Testing

Built and run on Linux/aarch64 against real eQSL, QRZ, Club Log and LoTW accounts. No automated tests: these hit live third-party services, and I did not want to add a suite that either needs credentials in CI or mocks the very reply formats that were the hard part. Happy to add tests over a fake HTTP layer if you would prefer them.

One unrelated note in case it appears in CI: `tests/scanner.rs::an_impossible_scan_is_refused_with_a_reason` fails for me, and fails identically on an unmodified `main`, so it is not from this branch.

### DIFF
--- a/crates/sdroxide-net/src/event.rs
+++ b/crates/sdroxide-net/src/event.rs
@@ -16,6 +16,10 @@ pub enum NetEvent {
     Callsign(CallsignInfo),
     /// One QSO/target upload result.
     Upload(UploadResult),
+    /// The outcome of checking one service's stored credentials. Distinct
+    /// from [`NetEvent::Upload`] because nothing was logged: this is an
+    /// answer about the account, not about a QSO.
+    LoginTest(sdroxide_types::LoginTestResult),
     /// Parsed confirmation records downloaded from LoTW/eQSL.
     Confirmations(Vec<QsoRecord>),
     /// WSPR reception reports fetched from WSPRnet — normally reports of *our*
--- a/crates/sdroxide-net/src/manager.rs
+++ b/crates/sdroxide-net/src/manager.rs
@@ -295,6 +295,35 @@ impl SpotManager {
         });
     }
 
+    /// Check one service's stored credentials; the result arrives via `poll`.
+    ///
+    /// On its own thread like every other network job, because these talk to
+    /// servers that can take seconds to answer and the UI thread is drawing a
+    /// waterfall. Read-only throughout: see [`crate::upload::test_login`].
+    ///
+    /// Tests the APPLIED config. The settings dialog therefore sends its edits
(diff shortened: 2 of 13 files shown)


dividebysandwich:
Looks good, thanks for the PR!

PR #109, opened 2026-08-20
Make the logbook window cheap to draw on a large log

FreespiritD:
Opening the Log window dragged the whole application down on a real logbook, with the waterfall visibly lagging for as long as the window stayed open and recovering the moment it closed. Measured on a Raspberry Pi 500 at 22,185 QSOs.

There were two separate costs, and this is both halves.

**Ordering and grouping, rebuilt every frame.** `log_list()` re-collected and re-sorted an index over the entire log on every frame, then re-derived every day boundary by calling `date_str` about twice per QSO, so roughly 44,000 heap allocations a frame at the size measured here. That alone cost 11.3 ms per frame, capping the UI at 88 fps before a single row was drawn. The spectrum shares that thread, which is why it is what suffers.

The ordering and grouping now live in a `LogView` refreshed only when the log has actually changed. The change check measures 0.124 ms at the same log size, about ninety times cheaper than what it replaces. Change is detected by hashing `id` and `start_utc` rather than by a generation counter: the log is mutated from import, delete, edit and the engine, and a site that forgets to bump a counter gives a silently stale list, which is a worse fault than the slowness being fixed. `id` catches an addition or deletion, `start_utc` catches an edit that moves a QSO to another day.

**Every row built every frame.** The list sat in a plain `ScrollArea`, which is not virtualised, so 22,185 QSOs constructed 22,185 frames, child Uis and labels per frame while about twenty were on screen. That was the larger of the two costs. The list is now flat, headers and rows in one `Vec<LogItem>` built alongside the grouping, and drawn through `ScrollArea::show_rows`. Per-frame cost stops depending on the size of the log.

**One visible change.** `show_rows` needs a uniform row height, so a day header now occupies the same vertical slot as a QSO row, which leaves a little more space around the date than before. The alternative, `show_viewport` with prefix-summed per-item offsets, keeps the header at its natural height and was rejected because its correctness rests on predicting egui's layout to the pixel for every item, and a few points of drift per item over tens of thousands of items is a scroll bar that lies about where it is.

**The row height is derived from the live style, not a constant.** `theme::apply_metrics` sets `item_spacing.y` to 5 on desktop and 7 on touch, and `interact_size.y` to 18 and 34, and re-applies both when the window crosses a breakpoint, so a row is 39 points on desktop and 53 on a phone tier. With a constant, resizing across that breakpoint would have silently desynchronised the scroll bar from the list.

**Testing.** `row_height_matches_the_live_style` lays a real row out headlessly in both tiers and asserts the formula matches what egui actually produced, so a margin or metric changed later fails a test rather than bending the scroll bar. Plus a test that the flattened list is a header followed by exactly that group's rows in order, and four tests on the caching: ordering and grouping including that the group ranges tile the ordering with no gap or overlap, that an unchanged log does not rebuild, that an addition, a deletion and a time edit each do, and the empty log. The whole `sdroxide-ui` suite passes at 346 tests, and `cargo test --release --no-default-features --no-run` is clean across the workspace.

**Not verified: how it looks.** These are layout changes and they want eyes on them, particularly the extra space around the day headers and the scroll bar when dragged the length of a large log.

### DIFF
--- a/crates/sdroxide-ui/src/app/logbook.rs
+++ b/crates/sdroxide-ui/src/app/logbook.rs
@@ -6,7 +6,7 @@
 //! asks them what would be new on every row of every slot.
 
 use eframe::egui::{self, Color32, RichText};
-use sdroxide_types::{LookupProvider, QsoRecord};
+use sdroxide_types::{LookupProvider, QsoRecord, UploadTarget};
 
 use crate::theme::ThemedScroll;
 use crate::time::now_unix;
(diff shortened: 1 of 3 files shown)


dividebysandwich:
Sorry I already took code from your branch!

PR #149, opened 2026-08-23
hpsdr: stop reporting RX discarded while keyed as a fault

reynico:
i've been catching this little guy for a while. 

the engine stops reading a receiver for the length of an over and drops the backlog on unkey, while the board keeps streaming RX so the RX ring fills and every datagram after that is discarded: expected, at the full RX rate, for as long as the operator transmits


dividebysandwich:
Looks good! Thanks a lot for the PR. This is gonna make a lot of HL2 owners happy. I'll probably adopt this for the other rigs as well.

PR #194, opened 2026-08-27
QO-100 satellite beacon sync offset and decode plugin

barisdinc:
when LNB or offset enabled in recever to work with QO-100 geostationay satellite , receiver locks to the beacon at 10489.750 Mhz, and also decodes the AO-40 telemetry from beacon and if succesfull decode is in place after automatically tuning the frequency for decode, it calculates the LNB or receiver offset error and corrects the frequency.
 Similar to QO-100 beacon plugin in SDR console

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5301,7 +5301,7 @@ dependencies = [
 
 [[package]]
 name = "sdroxide-qo100"
-version = "1.5.3"
+version = "1.5.4"
 dependencies = [
  "crossbeam-channel",
  "sdroxide-dsp",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5354,6 +5354,7 @@ dependencies = [
  "sdroxide-ism",
  "sdroxide-kiss",
  "sdroxide-net",
+ "sdroxide-qo100",
  "sdroxide-rigctld",
  "sdroxide-rotator",
  "sdroxide-skimmer",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5300,6 +5300,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "sdroxide-qo100"
+version = "1.5.3"
+dependencies = [
+ "crossbeam-channel",
+ "sdroxide-dsp",
+ "sdroxide-types",
+ "tracing",
+]
+
+[[package]]
 name = "sdroxide-r82xx"
 version = "1.5.4"
 dependencies = [
--- a/crates/sdroxide-proto/src/lib.rs
+++ b/crates/sdroxide-proto/src/lib.rs
@@ -827,7 +827,26 @@ use sdroxide_types::{
 /// Appended at the end of the struct; postcard numbers fields by position, so a
 /// v99 peer desynchronises on the tail of every `AdsbStatus` and the
 /// handshake's equality test is what stops it trying.
-pub const PROTO_VERSION: u16 = 100;
+///
+/// **101** — the QO-100 beacon calibration decoder.
+///
+/// [`sdroxide_types::RadioState`] gained `qo100`
+/// ([`sdroxide_types::Qo100Settings`]) holding whether the decoder runs and how
+/// wide it searches, and [`sdroxide_types::Command`] gained `SetQo100Config` to
+/// move it — the ISM decoder's shape, no apply step, the engine echoes the
+/// setting back in the state. Both are appended at the tail of their type, so
+/// every field and variant already on the wire keeps its number; a v100 peer
+/// desynchronises on the tail of every `RadioState` regardless, and the
+/// handshake's equality test is what stops it trying.
+///
+/// The live side — [`sdroxide_types::Qo100Status`] (lock, measured offset,
+/// decoded telemetry text) — is a native-engine `RadioEvent` only and never
+/// reaches the wire: `sdroxide-server` maps it to no `ServerMsg`, the same as
+/// the spectrum lanes that travel their own way. So there is no new `ServerMsg`
+/// here, and a remote client's window shows that the reading is local-only
+/// rather than sitting on "starting…". If it is bridged later that is its own
+/// append and its own bump.
+pub const PROTO_VERSION: u16 = 101;
 const VERSION_BYTE: u8 = 0x12;
 
 #[derive(Debug, thiserror::Error)]
--- a/crates/sdroxide-qo100/Cargo.toml
+++ b/crates/sdroxide-qo100/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "sdroxide-qo100"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+# Native-only: the QO-100 beacon decoder runs server-side in the engine, off
+# the raw IQ, like the skimmers and the ISM decoder. Not linked by the wasm
+# client.
+#
+# The BPSK-400/AO-40-uncoded framing implemented here is written from public
+# protocol facts (sync word, frame length, CRC variant, encoding order) — not
+# ported from any existing decoder's code — cross-checked against the open
+# reference implementation in Daniel Estévez's gr-satellites project
+# (GPL-3.0-or-later, https://github.com/daniestevez/gr-satellites,
+# python/components/deframers/ao40_uncoded_deframer.py and
+# python/satyaml/QO-100.yml). See `bpsk.rs` for the citations inline.
+
+[dependencies]
+sdroxide-types.workspace = true
+sdroxide-dsp.workspace = true
+crossbeam-channel.workspace = true
+tracing.workspace = true
--- a/crates/sdroxide-qo100/src/bpsk.rs
+++ b/crates/sdroxide-qo100/src/bpsk.rs
@@ -1,0 +1,577 @@
+//! BPSK-400 demodulation and AO-40 "uncoded" frame decode for the QO-100
+//! narrowband beacon.
+//!
+//! # Protocol, confirmed rather than guessed
+//!
+//! Every number below is taken from Daniel Estévez's (EA4GPZ) gr-satellites
+//! project (GPL-3.0-or-later, <https://github.com/daniestevez/gr-satellites>)
+//! — he wrote and maintains the QO-100 beacon decoder there — cross-checked
+//! against his write-up at
+//! <https://destevez.net/2017/05/decoding-ao-40-uncoded-telemetry/>. Nothing
+//! here is ported from that code; these are protocol facts (a sync pattern,
+//! a frame length, a CRC variant), not an implementation.
+//!
+//! - `python/satyaml/QO-100.yml`: 10489.750 MHz, "DBPSK Manchester", 400 baud,
+//!   framing "AO-40 uncoded" (the satellite also sends an FEC-coded variant on
+//!   alternate frames; this module does not attempt that one — see the crate
+//!   doc).
+//! - `python/components/deframers/ao40_uncoded_deframer.py`: sync word
+//!   `'00111001000101011110110100110000'` (32 bits, MSB first —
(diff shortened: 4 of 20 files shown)


dividebysandwich:
Thanks for this PR, that's a great feature! However I am not going to merge it as it is. There are three things that block it outright and one design problem in the search.

Blocking:
"cargo check --workspace --all-targets" fails:

error[E0063]: missing fields rows and rows_clocked in initializer of SpectrumFrame --> crates/sdroxide-ui/src/app/qo100.rs:611

The test helper frame_with was written against a SpectrumFrame from before the waterfall row-clock change. A lib-only build passes, which is why this got through, but "cargo test -p sdroxide-ui" cannot build. With that one line fixed locally, all 455 sdroxide-ui tests and all 9 sdroxide-qo100 tests pass, so it is the only test breakage. Please rebase to the latest master.

PROTO_VERSION is not bumped. RadioState gains a qo100 field and Command gains SetQo100Config, but crates/sdroxide-proto/src/lib.rs is untouched. Postcard is positional and not self-describing, and ServerMsg::HelloAck carries rx_codec and tx_codec after state, so a v96 client and this server silently disagree about every state message. The version history in that file documents this same rule for every previous RadioState change (v63 for the ISM block, v95 for SetDigiTxLevel). There is no golden-bytes test, so nothing in CI catches it. Please bump to the current PROTO_VERSION + 1 with a history entry alongside the others.

Clicking OFF, or the search width buttons, can freeze the engine for minutes. Qo100Controller's Drop joins the worker, and the worker's unit of work is one whole bpsk::acquire. The engine thread drops that controller in sync_qo100 (when switching OFF), in sync_qo100_window (on any width or sample-rate change), and on device swaps. Measured acquire times, release build:

+/-1 kHz    demod rate 16 kHz     0.81 s
+/-5 kHz    demod rate 16 kHz     3.95 s   (this is the default)
+/-10 kHz   demod rate 25 kHz    11.40 s
+/-25 kHz   demod rate 62.5 kHz  66.74 s
+/-50 kHz   demod rate 125 kHz  264.50 s   (this is MAX_HALF_WIDTH_HZ)

The budget is 11.9 s of audio per window. So a width click at the top setting stalls receive and transmit for about four and a half minutes, and even the default costs a four second dropout. The pattern is lifted from SkimmerController, where the join is free because its unit of work is one block's FFT. Here it needs a stop flag acquire can check, or a detached worker.

There's also a problem with the search design: Cost grows quadratically with the search width, and the UI offers widths that can never keep up. qo100_target_rate_hz raises the demodulation rate along with the search width (half_width * 2.5) while the candidate count grows with it as well. The wider capture bandwidth is genuinely needed, but the demodulator only wants about 16 kHz no matter how wide that capture is, so each candidate should be mixed down to a fixed demod rate rather than searched at 125 kHz. As written, anything at or above +/-10 kHz is over real time, and the status line's "first block takes about 10 s" is off by a factor of 25 at the maximum.

Dropped IQ blocks splice the coherent window. on_rx_iq try_sends into a 64-deep channel, which is about half a second of audio at a 16384-sample device read, and the worker extend_from_slices whatever survives. Above +/-10 kHz backpressure is the steady state, not an edge case, so a hole gets punched into a buffer that a 10.36 s frame has to sit inside contiguously. This is the same trap as the DeepCW contiguous-window rule. What saves it today is only that the kept 11.9 s prefix is still longer than one frame; the comment saying a dropped block "just gets tried again in the next window" understates it.

can_rx_hz is the wrong reachability gate. qo100.rs:279 is the only non-test use of can_rx_hz for gating anywhere in the tree; everything else that gates behaviour uses may_rx_hz. On a driver that publishes no receive ranges (SoapySX among others, which may_rx_hz's own doc calls out), a correctly configured LNB station gets ON greyed out and is told to set up an LNB/converter offset that it already has. readout_digits has the same blind spot: empty ranges means 10 columns, so the dial truncates 10489.750 MHz, which is the exact bug the extra column was added to prevent.

Minor issues:

There is no docs/USER_MANUAL.md section, look at ISM and the skimmers for a template.

qo100_cfg is written and never read. Its comment says it restores what was chosen across an audio-mode swap, but nothing does that. ism_cfg has the same shape, so this is a copied pattern rather than a new problem, but the comment should not claim behaviour that is not there.

Every test uses a +/-50 to 150 Hz search and finishes in 0.06 s, which is why the cost blowup went unnoticed. Something should pin the search cost at a realistic width.

False lock arithmetic: roughly 110 chance sync hits per search at the default width, times 2^-16 for the CRC, works out to about one false lock per couple of hours of searching. APPLY writes that straight into converter_offset_hz. The telemetry text panel is a real mitigation since garbage would be visible, but nothing gates APPLY on it.

RadioEvent::Qo100Status(_) => None in the server means a remote or web client's window reads "starting..." forever. That is documented as deliberate and I am fine with it for now, but the window should say so rather than looking stuck.


Again, please don't take this as a rejection since this is a great contribution, and I hope you could address the points above so we can merge this in!

barisdinc:
Thannks for the comment.. I will sync to HEAD and take care of the issues you mentioned ASAP...
thanks
Baris
TA7W/OH2UDS

barisdinc:
Thanks for the detailed review — this was exactly the kind of pass the feature needed. The branch **is now rebased** onto current main and every point below is addressed, one focused commit per item. Replying inline to each part of your comment.

Blocking

**▎ cargo check --workspace --all-targets fails: error[E0063]: missing fields rows and rows_clocked in initializer of SpectrumFrame --> crates/sdroxide-ui/src/app/qo100.rs:611 … The test helper frame_with was written against a SpectrumFrame from before the waterfall row-clock change … Please rebase to the latest master.**

Fixed in 24b839a. frame_with in crates/sdroxide-ui/src/app/qo100.rs now sets rows / rows_clocked (helper at qo100.rs:647). Branch rebased onto current main (merge-base is now main@HEAD). cargo check --workspace --all-targets is clean on the branch — no errors; the only warnings are pre-existing lints in sdroxide-adsb / sdroxide-skimmer / sdroxide-ism test code, untouched by this PR.

**▎ PROTO_VERSION is not bumped. RadioState gains a qo100 field and Command gains SetQo100Config, but crates/sdroxide-proto/src/lib.rs is untouched. Postcard is positional and not self-describing … Please bump to the current PROTO_VERSION + 1 with a history entry alongside the others.**

Fixed in a30e6a5. PROTO_VERSION 100 → 101 (crates/sdroxide-proto/src/lib.rs:849), with a history paragraph in the same style as the v63 (ISM) and v95 (SetDigiTxLevel) entries, noting both tail-appended additions. The paragraph also records why there is no new ServerMsg: RadioEvent::Qo100Status maps to None in sdroxide-server, so the live side never reaches the wire (this is the "starting…" point below).

**▎ Clicking OFF, or the search width buttons, can freeze the engine for minutes. Qo100Controller's Drop joins the worker, and th▎ worker's unit of work is one whole bpsk::acquir needs a stop flag acquire can check, or adetached worker.**

Fixed in 150e909. Qo100Controller now carries a s (crates/sdroxide-qo100/src/controller.rs:92).bpsk::acquire polls it between candidates (bpsk.rs:388) and returns early; Drop sets it before the join (controller.rs:237), so a drop waits out at most one candidate instead of a whole sweep. This covers all three drop sites you listed (sync_qo100, sync_qo100_window, device swap).

Search design
 **▎ Cost grows quadratically with the search width about 16 kHz no matter how wide that capture is,so each candidate should be mixed down to a fixed demod rate rather than searched at 125 kHz … the status line's "first block takes about 10 s" is off by a factor of 25 at the maximum**

Fixed in 150e909. acquire now takes an explicit demod_rate_hz (~16 kHz, bpsk.rs:375) and does a combined mix-and-decimate to that rate per candidate before the chip search, so per-candidate cost no longer tracks the capture rate. Candidates are tried centre-outward, so a roughly-calibrated station is found before the grid is walked in full. The status line's stale "~10 s" is corrected to the actual ~24 s default-width window. Regression guards in d5b55f8: a search pinned at a realistic capture rate +the engine default ±5 kHz width to a 5 s budget ( search that returns without walking the grid(bpsk.rs:602).

**▎Dropped IQ blocks splice the coherent window. on_rx_iq try_sends into a 64-deep channel … above +/-10 kHz backpressure is the steady state … a hole gets punched into a buffer that a 10.36 s frame has to sit inside contiguously.** 

Fixed in 150e909. The IQ channel is widened to 256, and the worker now tracks whether the realtime side had to drop a block; when it did, the rolling buffer restarts from the next contiguous block instead of extend_from_slice-ing across the gap. Fewer windows under sustained backpressure, but every of them is whole air.

**▎ can_rx_hz is the wrong reachability gate … everything else that gates behaviour uses may_rx_hz … eadout_digits has the same blind spot: empty ranges means 10 columns, so the dial truncates 10489.750 MHz.**

Fixed in 96f3df3. The ON gate moves to may_rx_hz (crates/sdroxide-ui/src/app/top_bar.rs:4879), matching every other reachability check in the tree, so a driver that publishes no RX ranges no longer greys out a correctly configured LNB station. readout_digits now also earns the extra column from the configured converter offset and the dial frequency itself, independently of the range list, so 10489.750 MHz is never truncated.

Minor issues

**▎ There is no docs/USER_MANUAL.md section, look at ISM and the skimmers for a template.**

f2f8b28 — added §2.21 on the ISM/skimmer templatek-safe APPLY behaviour, and the remote/browserlocal-readout note.

**▎ qo100_cfg is written and never read. Its commenhosen across an audio-mode swap, but nothing doesthat … the comment should not claim behaviour that is not there.**

656c307 — rewrote the comment in crates/sdroxide-be what actually happens (the decoder stays off after an audio-mode swap until the operator turns it back on). Left ism_cfg's identical wording alone as out of scope, but the note now flags it as the same copied pattern.

**▎ Every test uses a +/-50 to 150 Hz search and finishes in 0.06 s, which is why the cost blowup went unnoticed. Something should pin the search cost at a realistic width.**

Covered by the bpsk.rs:588 budget test above, plus d5b55f8 adds a controller::tests module that exercises the worker thread, the rolling buffer and the status channel end to end (synth_signal is now pub(crate) so it can build a real on-air frame):    freshness of lock_is_fresh, window overlap > 1 froll snapshot, and a synthesized frameround-tripping through the worker to a decoded lock.

**▎ False lock arithmetic: … about one false lock pites that straight into converter_offset_hz …nothing gates APPLY on ....** 

Fixed in 14a71bc. APPLY CORRECTION now stays disa frame lands (blocks_locked >= 2) carryingnon-empty decoded text (apply_is_confirmed, crates/sdroxide-ui/src/app/qo100.rs:578). A single chance sync+CRC hit no longer enables the button; the hover text explains the wait rather than just greying out. Tested against one lock, two blank-payllocks, and two text-bearing frames.

**▎ RadioEvent::Qo100Status(_) => None in the server means a remote or web client's window reads "starting…" forever … the window should say so rather than looking stuck.**

Fixed in f897649. status_line now takes engine_is_remote() (qo100.rs:435) and, for a remote/browser client, states plainly that the readout is local to the engine host instead of showing a status that will never arrive.

9ed6617 pulls the three decisions that were inline in impl blocks out into free, tested functions so their edge cases are pinned: qo100_capture_rate_for (engine.rs:2317, ttonicity), apply_is_confirmed (qo100.rs:578), andreadout_digit_count(range_reaches_10ghz, converter_offset_hz, active_freq_hz) (top_bar.rs:5039, HF/VHF/23 cm vs 13 cm transverter vs QO-100 LNB vs dial-on-beacon).

Force-pushed as a clean 14-commit series on top of current main. Happy to squash or reorder if you'd prefer a different shape for the merge.

Thanks
Baris
OH2UDS/TA7W

dividebysandwich:
Excellent, thanks a lot! A handful of minor things, but I'll do them after the merge. This is good stuff!!

PR #274, opened 2026-09-01
Add RigExpert FobosSDR support

### LINKED ISSUE #273: Feature Request: Support for FobosSDR
The FobosSDR is an SDR made in Ukraine by RigExpert that covers from 100 kHz to 6 GHz. This is divided into HF, which is direct sampled, and RF, which is tuned before being sampled. HF delivers two IQ streams from two antennas, making the receiver a cost effective way to explore phasing, and the 14-bit ADC makes it more effective than some other radios that offer this functionality.

thereisnocat:
Closes #273.

A native backend for the FobosSDR: RF port (tuner), HF1/HF2 direct
sampling, and HF1+HF2 combined by an adaptive diversity filter for
noise cancellation or maximal-ratio combining. The FobosSDR is a
reasonably priced SDR covering from 100 kHz to 6 GHz made by RigExpert
in Ukraine. Its architecture is similar in many ways to the rx888, at
least in the HF direct sampling part of the receiver.

## Driver

crates/sdroxide-fobos wraps libfobos (LGPL-2.1) via dlopen at runtime,
the same pattern this fork's SDRplay and LimeSDR backends already use
-- nothing linked at build time, so every build variant ships with
this backend and a machine without the vendor library simply
enumerates nothing.

## HF1/HF2 direct sampling

The FobosSDR has no local oscillator on its HF ports: fobos_rx_read_sync
hands back two independent real ADC channels (HF1 in one, HF2 in the
other -- confirmed by reading libfobos's own source). Each channel is
turned into complex baseband with sdroxide_dsp::WbDdc, the same wideband
real-to-complex channelizer sdroxide-rx888 already built and proved for the
identical direct-sampling problem.

The raw hardware read and the DDC's own FFT compute run on separate
threads (mirroring sdroxide-rx888's own split): on real hardware,
running them back to back on one thread left the DDC compute
delaying the next read by nearly its own duration again, audible as
frequent dropouts.

## Diversity (HF1+HF2)

Combining uses sdroxide_dsp::Diversity, the adaptive NLMS filter used
for the RSPduo and LimeSDR's second chain. Verified live: 38.6 dB of
real cancellation on real antennas, stable across repeated measurements.

## Verified live, repeatedly, against real hardware

Every phase of this backend -- RF-port streaming, HF-port direct
sampling, dual-channel diversity, and the bug fixes below -- was
checked against a real FobosSDR, not just built and assumed working.
Notable real-hardware findings along the way:

- The reference rate for direct-sampling streaming isn't the ADC's
  own top rate: requesting a lower rate is honored (down to an
  8 Msps floor below which requests snap up regardless), and running
  the ADC no faster than the request actually needs fixed the
  DDC-compute throughput problem above.
- Two ADC rates were found broken specifically for *streaming* (not
  for the samplerate-set call, which reports success either way):
  40 Msps never streams at all, and 50 Msps streams but produces
  audible distortion on a real listen. The rate-selection logic
  steers requests away from both, verified against several nearby
  candidates live before picking a safe one.
- HF targets above roughly 8 Msps reintroduce the DDC-compute
  ceiling regardless of ADC-rate tuning, since a wide view genuinely
  needs a fast ADC -- documented as a known, likely-architectural
  limit rather than worked around with another constant.
- The very first read after starting a stream can fail on a startup
  race, independent of which rate is in use; retried automatically
  now rather than treated as fatal.
- HF1/HF2 have no LO, so the receivable ceiling is always half
  whatever rate the ADC is actually running at -- a real tuning-range
  bug (2.5 Msps left 10 MHz as the practical ceiling) that a
  same-session attempt to fix by simply raising the ADC rate had to
  be reverted after it reintroduced the same audible distortion.
- The diversity filter's own Restart control could zero the adaptive
  filter without clearing its own Hold flag, silently leaving Cancel
  mode parked at zero (indistinguishable from Combine) until Hold was
  separately toggled -- fixed in this backend specifically; the
  identical gap exists in the shared Diversity code the RSPduo and
  LimeSDR backends also use, worth a reviewer's eye on whether the
  real fix belongs there instead.

cargo build/test/clippy clean throughout.

This code was created with the assistance of Claude Code, with full 
testing against actual hardware and revisions by Ralph Brandi.


dividebysandwich:
Excellent thank you! I did some minor touch-ups, didn't want to make you wait for another round.

Issue #301, opened 2026-09-03
CW offset


parlotto:
I use sdroxide on linux with a redpitya in HPSDR compatible mode and a conventional TS590 hardware radio.
 I noticed a offset in CW mode between sdroxide and TS590. The offset appears only in CW mode. In SSB dial frequencies match perfectly on the two radio. 
In LSB for example the red line stand for carrier frequency and the audio pass-band  is highlighted below the carrier which is the display used by most SDR programs :
<img width="513" height="735" alt="Image" src="https://github.com/user-attachments/assets/fdafd54a-8527-460e-a0e8-7745a285cf0b" />
In CW the carrier is displayed below the pass-band  and therefore the dial frequency is not the carrier frequency (about 700Hz lower which is  I think the tone frequency offset)
<img width="415" height="575" alt="Image" src="https://github.com/user-attachments/assets/12110508-0d15-4707-9092-58462c1e74e4" />
Here is the display with GQRX and the same hardware : audio pass-band is correctly centered on the carrier and perfectly match with my hardware radio :
<img width="505" height="375" alt="Image" src="https://github.com/user-attachments/assets/6df7d31e-b891-42b5-a043-b09c25557223" />
Here is my ts590 transmitting CW on **7.002.00** kHz , I have to tune 700Hz below to center the signal in the pass-band : 

<img width="407" height="677" alt="Image" src="https://github.com/user-attachments/assets/42695b8a-89bf-44e1-8476-b84b3694de58" />

This is annoying because the dial CW frequency is not the carrier frequency and because I want synchronize the two radio with hamlib server. 
Is there a way to modify the CW offset ?  Can the dial frequency be equal to carrier frequency is CW ?

dividebysandwich:
Hmm, I think this could be a user setting.

dielectric-coder:
First, since it is my first comment here, thanks for this absolutely wonderful application, i played with them all and this one is a keeper for me ! :)
[...]

PR #304, opened 2026-09-03
Add a QRG chip to the CW panel, so the readout shows the frequency being worked rather than the dial

dielectric-coder:
A CW dial sits a sidetone pitch below what is being copied — 700 Hz, usually — so the big readout has never been the frequency of the contact. It is that frequency minus the pitch, and the subtraction is left to the operator, every time, on a number they are about to put in the log or say on the air.

This adds a **QRG** chip to the CW panel that puts the main readout and the tuning line on the signal instead.

## Relation to #301 — partial

refs #301. That report describes this exactly: in CW the carrier sits *below* the highlighted passband, so the dial reads about 700 Hz low against a TS-590 on the same signal, where in LSB the two agree. With the chip on, the readout shows the carrier and the tuning line moves into the middle of the passband, which is what the screenshots there are asking for and what GQRX is shown doing.

**It does not close the issue**, for two reasons worth stating plainly:

- The dial itself is unchanged. This moves what is *displayed*, not where the receiver is tuned, so the answer to "can the dial frequency be equal to the carrier frequency in CW" is still no — the offset is now shown rather than removed.
- Because of that, **Hamlib is unaffected**: `rigctld` reports `RigState::vfo_a_hz`, the raw dial, so synchronising sdroxide with another radio over rigctld will still be a sidetone out. That is specifically what the reporter wants it for, so this PR does not solve their problem end to end.

Making the dial itself the carrier is a larger change — it moves the demodulator's cursor, memories, band-edge checks and what rigctld publishes, all of which currently agree that the dial is the dial. Worth doing, probably, but it is a different PR and wants its own discussion.

## Most of it was already there

I expected this to need more than it did. `Mode::on_air_hz` has always answered the real frequency, the CW panel has always displayed it, and `Engine::sync_cw_filter` has always kept the passband centred on the pitch. What was left on the dial was the **main readout** and the **tuning line** — so the red line sat at the *edge* of its own passband, and the two frequencies on screen disagreed by a sidetone.

So the change is small and entirely in the presentation layer:

- **The readout** goes through the offset `readout_for` already returns for a repeater shift. That is what keeps it tunable: a wheel turn or a typed frequency still moves the dial by what the operator changed rather than jumping it by the pitch. A keyed repeater shift still outranks it — where RF is actually going matters more — and there is a test for that precedence.
- **The tuning line** rides a new flag on the `AudioCursor` the panadapter already draws at the pitch, so it lands where the passband is already centred. Line, shading and signal finally agree.

## Why "QRG"

The Q-code for "your frequency is" — which is exactly the question a CW dial cannot answer on its own, and exactly the number this puts in the readout. It names the thing rather than the switch: an operator reads a label to find out what the number will mean, not to learn that a mode has been turned on. The doc comments spell the code out for anyone who does not use it.

## Scope

Display only. The dial is unchanged and everything that tunes, stores or checks a band edge goes on using it. Off by default, which is what every other radio does, and persisted per screen in `UiSettings` rather than on the wire — this is a preference about reading, not about the radio, so **no `PROTO_VERSION` bump**.

**CW alone for now.** RTTY and WEFAX sit off their dials too (`Mode::tunes_off_dial` already covers them) and the same treatment would suit them, but each wants checking against real signals first — WEFAX's offset is large enough that the tuning line would move visibly.

## Also

Removes a third copy of the hardcoded 700 Hz pitch fallback. The readout asks a new `cw_pitch_hz()` helper that follows the controller's live cursor exactly as the engine's own does, so the readout and the passband cannot disagree about where the signal is.

## Testing

- Three new tests: the readout follows the *live* pitch rather than a fixed 700, agrees with `Mode::on_air_hz`, is inert outside CW, and loses to a repeater shift when keyed.
- `cargo test --workspace` green apart from `sdroxide-radio --test repeater`, which fails on a clean `main` too and is untouched here.
- Exercised in **both** clients, not just compiled for them: the native GUI, and the browser client served by `sdroxide --server` from an `embed-web` build. The chip behaves the same in each, which is the part worth checking on a UI crate that targets wasm as well as the desktop.
- **Checked against a signal generator on an RSPdx**: a carrier of known frequency tuned in CW reads that frequency in the main readout with the chip on, and the tuning line sits in the middle of the passband rather than at its edge.

One interaction worth knowing, and called out in the chip's hover text: click-to-tune rounds the clicked frequency to `click_tune_step_hz` (10 Hz by default) *before* the pitch is subtracted, so a coarse step leaves the signal off the cursor by up to half of it. That was always true; QRG just makes it visible, because the readout is now claiming a precise number.

## LLM usage

Written with LLM assistance, in line with the project's guidelines. I have read and reviewed the whole diff, it is tested against real hardware as described above, and I can answer questions about any part of it.

### DIFF
--- a/crates/sdroxide-types/src/ui.rs
+++ b/crates/sdroxide-types/src/ui.rs
@@ -538,6 +538,25 @@ pub struct UiSettings {
     /// Decode-list filter: only stations that would put something new in the
     /// log (new entity, new band-slot, new grid, or a callsign never worked).
     pub decode_new_only: bool,
+    /// In CW, read the frequency of the *signal* rather than of the dial —
+    /// the QRG, in the Q-code an operator would use to ask for it.
+    ///
+    /// A CW dial sits a sidetone pitch below what is being copied, so the
+    /// number in the readout is not the number either operator would quote —
+    /// it is that number minus the pitch, and the arithmetic is left to the
+    /// person. With this on, the main readout shows what
+    /// [`crate::Mode::on_air_hz`] answers, and the tuning line on the
+    /// panadapter moves to sit on the signal, where the passband is already
+    /// centred (the engine keeps the filter on the pitch).
+    ///
+    /// A display preference, not a change of tuning: the dial is still the
+    /// dial, and everything that tunes, stores or checks a band edge goes on
+    /// using it. Off by default, which is what every other radio does.
+    ///
+    /// CW only for now. RTTY and WEFAX sit off their dials too
+    /// ([`crate::Mode::tunes_off_dial`]) and could join it, but each wants
+    /// checking against real signals first.
+    pub cw_qrg: bool,
 }
 
 /// Default for [`UiSettings::spot_colors`] — every kind on its stock tint.
@@ -596,6 +615,10 @@ where
 impl Default for UiSettings {
     fn default() -> Self {
         UiSettings {
+            // Off: the dial is what every other radio shows, and an operator
+            // who has not asked for the change should not find their readout
+            // reading differently from the rig beside it.
+            cw_qrg: false,
             frame_rate_fps: 60,
             waterfall_speed: Speed::Medium,
             spectrum_speed: Speed::Medium,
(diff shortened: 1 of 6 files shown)


dividebysandwich:
Landed in a0c081a0

PR #370, opened 2026-09-06
SpyServer: follow the dial with the FFT window instead of the server's start-up frequency

### LINKED ISSUE #369: SpyServer: tuning does nothing — the receiver stays on the frequency the server started on
## Symptom

Against a SpyServer publishing an Airspy HF+, the receiver never leaves the frequency the **server** was started on. Tuning the dial anywhere — 7.1 MHz, 14.1 MHz, a different band entirely — changes the labels on screen but not the audio and not the spectrum: what arrives is always the server's own start-up frequency, 100 MHz in my case.

Nothing in the log looks wrong. sdroxide reports the centre it was asked for:

```
SpyServer 10.90.250.100:5555: Airspy HF+ (serial 39333038), 0.768 Msps max, 16-bit,
  gain 0..0, 0.000-1700.000 MHz - 8-bit (lowest bandwidth) I/Q at 768.000 ksps
  (stage 0, 12.29 Mbit/s on the link), plus a 0.660 MHz FFT for the full-band strip
INFO sdroxide::spyserver_source: SpyServer source ready: SpyServer 10.90.250.100:5555
  - Airspy HF+, 768.000 ksps 8-bit (lowest bandwidth), 12.3 Mbit/s, center 7100000 Hz
INFO sdroxide_radio::engine: skimmer started rate=192000.0 center=7100000.0
```

The same server tunes correctly from SDR++, which is what sent me looking at the client rather than at the server.

## What is actually happening

With the FFT lane running, this server tunes its **receiver** to `SETTING_FFT_FREQUENCY`. `SETTING_IQ_FREQUENCY` only positions the I/Q window *inside* the band the receiver is already on — and at I/Q decimation stage 0 there is no room to position it at all, so it does nothing.

Measured directly against the server with a minimal protocol client, capturing I/Q and comparing each spectrum against references taken in IQ-only mode at known frequencies:

| what was set | where the receiver actually was |
| --- | --- |
| I/Q 7.1 MHz, FFT 100 MHz (what sdroxide sends) | 100 MHz (corr +0.995) |
| the same two values, I/Q sent **last** | 100 MHz (corr +0.994) — so it is not last-write-wins |
| FFT lane on, then only `IQ_FREQUENCY` retuned to 7.1 | 100 MHz (corr +0.991) |
| both windows moved to 7.1 MHz | 7.1 MHz (corr +0.885) |
| open both on 7.1, then move **only** the FFT window to 14.1 | 14.1 MHz (corr +0.919) |
| open both on 7.1, then move **only** the I/Q window to 14.1 | still 7.1 MHz (corr +0.932) |

SDR++ is unaffected because it streams IQ-only and never sends `FFT_FREQUENCY`; in that mode `IQ_FREQUENCY` does move the receiver.

## Root cause

`Client::plan` opens the FFT window at `clamped_fft_center(self.center)`, and that clamps into `device_center ± fft_slack_hz(span)`:

- `device_center` is whatever the handshake's `CLIENT_SYNC` reported — where the server was sitting *before* we tuned it.
- `fft_slack_hz` at FFT decimation stage 0 is `(maximum_bandwidth - maximum_bandwidth) / 2` — exactly **zero**, because the window already covers the whole receiver.

So the clamp collapses to precisely the server's own frequency, whatever the operator asked for, and `configure()` sends it. The receiver follows the window and parks there.

It never recovers either. `maintain_fft` re-derives the target through `DeviceInfo::fft_recenter`, which clamps against the same `device_center` — and this server sends exactly one `CLIENT_SYNC`, at the handshake, so `device_center` stays 100 MHz for the life of the session. Every pass recomputes the same answer.

Stage 0 is the default (`SpyServerConfig::fft_decimation: 0`, `fft_enabled: true`), so any server whose start-up frequency is not where you want to listen shows this.

## Environment

- sdroxide 1.6.5, built from `f3af773`
- macOS 14 (aarch64), `cargo build --release --no-default-features --features rtl433`
- Server: SpyServer publishing an Airspy HF+ — 768 ksps, 660 kHz analog bandwidth, `CanControl = 1`, gain index 29, start-up frequency 100 MHz

I have a fix and will open a PR shortly. It is tested against the same real receiver, plus regression tests against a fake server shaped like this one.

drfruct:
Fixes #369.

## The bug

Against a SpyServer publishing an Airspy HF+, tuning did nothing: whatever frequency the dial was set to, the receiver stayed on the frequency the **server** had started on. The log reported the requested centre, so nothing looked wrong from this side.

## Root cause

With the FFT lane running, a SpyServer tunes its **receiver** to `SETTING_FFT_FREQUENCY`. `SETTING_IQ_FREQUENCY` only positions the I/Q window _inside_ the band the receiver is already on — and at I/Q decimation stage 0 there is no room to position it, so it does nothing at all.

`Client::plan` opened the FFT window at `clamped_fft_center(self.center)`, which clamped into `device_center ± fft_slack_hz(span)`:

- `device_center` is whatever the handshake's `CLIENT_SYNC` reported, i.e. where the server sat _before_ we tuned it;
- `fft_slack_hz` at FFT stage 0 is `(maximum_bandwidth - maximum_bandwidth) / 2`, exactly **zero**, because the window already covers the whole receiver.

The clamp therefore collapsed to precisely the server's own frequency, whatever the operator asked for, and `configure()` sent it. The receiver followed the window and parked there.

It never recovered, either: `maintain_fft` re-derived the target through `DeviceInfo::fft_recenter`, which clamped against the same `device_center` — and the server I hit sends exactly one `CLIENT_SYNC`, at the handshake, so `device_center` stayed put for the life of the session and every pass recomputed the same answer.

Stage 0 is the default (`fft_decimation: 0`, `fft_enabled: true`), so this shows on any server whose start-up frequency is not where you want to listen. SDR++ is unaffected because it streams IQ-only and never sends `FFT_FREQUENCY`; in that mode `IQ_FREQUENCY` does move the receiver.

## Measured on the wire

Captured I/Q from a real Airspy HF+ over SpyServer with a minimal protocol client and correlated each spectrum against references taken in IQ-only mode at known frequencies:

| what was set                                                | where the receiver actually was             |
| ----------------------------------------------------------- | ------------------------------------------- |
| I/Q 7.1 MHz, FFT 100 MHz (what sdroxide sent)               | 100 MHz (corr +0.995)                       |
| the same two values, I/Q sent **last**                      | 100 MHz (corr +0.994) — not last-write-wins |
| FFT lane on, then only `IQ_FREQUENCY` retuned to 7.1        | 100 MHz (corr +0.991)                       |
| open both on 7.1, then move **only** the FFT window to 14.1 | 14.1 MHz (corr +0.919)                      |
| open both on 7.1, then move **only** the I/Q window to 14.1 | still 7.1 MHz (corr +0.932)                 |
| both windows moved together to 7.1                          | 7.1 MHz (corr +0.885) — the fixed behaviour |

A further measurement decided the ordering below: with a decimated I/Q window, moving the FFT window carries the receiver but leaves the I/Q window on its old absolute frequency, so a position sent against the outgoing band stays there.

## The fix

- `DeviceInfo::fft_recenter` now takes whether this client controls the receiver, and how much slack the I/Q window has. Where it does control it, the window is the receiver: it recentres on the dial, and it may only hold still while the I/Q window can still reach the dial on its own. At the top of the rate ladder that slack is zero, so the window tracks the dial exactly. Where another client owns the receiver, nothing changes — the device does not move for us, the window slides inside it, and the hysteresis runs its full width.
- `Client::clamped_fft_center` no longer clamps against the device centre when this client controls the receiver. That clamp was the bug.
- The FFT window is now placed **before** `IQ_FREQUENCY`, both at connect and on every retune, because the I/Q position is only meaningful once the receiver is where it belongs. `maintain_fft` keeps its rate limit for the case it exists for — a shared server's owner moving the device — while a retune moves the window directly, having already paid that interval.
- Switching the FFT lane on mid-session recomputes the window centre instead of sending whichever one it was left holding, which would otherwise drag the receiver back to the band the operator was in when the lane was last on.

## Testing

Per the contributing rules, this was tested against real hardware: an Airspy HF+ published over SpyServer, tuning across 40 m, 20 m and the FM broadcast band, confirmed both by ear and by the correlations in the table above.

Automated coverage added:

- four tests in `net.rs` against a fake server shaped like the one this was found on — an HF+ whose analog bandwidth is its whole stage-0 FFT span, sitting on 100 MHz, sending exactly one `CLIENT_SYNC`. Three of them fail on the current `main`;
- two `fft_recenter` unit tests for the controlled-receiver case, covering both the zero-slack and the still-has-slack halves;
- the existing shared-server tests are kept and now guard the other half of the branch.

`cargo test -p sdroxide-spyserver` is green (60 tests), as are the root crate's four `spyserver_source` tests. No new clippy warnings; `cargo fmt` clean.

One existing assertion changed. `the_fft_block_is_configured_before_the_streaming_mode` asserted that the FFT block follows `IQ_FREQUENCY` — which is the ordering being fixed. It now asserts that the block follows the I/Q lane's _shape_ (format and decimation) and that `FFT_FREQUENCY` precedes `IQ_FREQUENCY`, with the reason written down.

## Second fix in the same PR: the HF+'s 8-bit digital gain was never sent, and was too small when it was

Found the moment tuning worked: on a quiet band (1.08 MHz, next to an NDB) the panadapter showed a flat noise floor across the whole 768 kHz with a deep, ripple-edged dip only around the dial — as if everything but the tuned frequency were 20 dB noisier. The server's own FFT strip of the same band was flat and clean, so it was the I/Q lane.

### Measured on the wire

Raw I/Q captured off the socket at 1.0814 MHz, before any client DSP, floor by distance from the centre (dBFS/bin, median per band) against a 16-bit capture of the same band. The last column is the strongest sample seen, out of 127.

| stream | 0–20 kHz | 150–250 | 250–330 | 330–384 (receiver roll-off) | distinct values | max |
| --- | --- | --- | --- | --- | --- | --- |
| 16-bit, 0 dB (reference) | -83.1 | -83.9 | -83.3 | -94.4 | — | — |
| 8-bit, 0 dB (what sdroxide sent) | -78.6 | -60.1 | -59.8 | — | **3** | 1 |
| 8-bit, 32 dB (what it meant to send) | -83.0 | -80.9 | -79.6 | -80.7 | 11 | 5 |
| 8-bit, 38 dB | -83.1 | -83.0 | -82.0 | -86.1 | 15 | 7 |
| **8-bit, 44 dB** | -83.0 | -83.9 | -83.3 | -91.3 | 41 | 20 |
| 8-bit, 50 dB | -83.0 | -83.9 | -83.3 | -95.6 | 78 | 41 |

At 0 dB the HF+'s I/Q sits below one 8-bit LSB, so the server ships a three-level slicer output: a flat floor 23 dB above the receiver's own, with the dip at DC that is the shape of that quantisation error rather than any filter.

At 32 dB the noise is ±2 LSB and the quantiser is still a nonlinearity rather than a noise source: the floor rises 3.4 dB towards the band edges and fills the receiver's roll-off 14 dB above the truth, which shows on the panadapter as a bathtub — quiet in the middle, louder at both ends, where SDR++ (16-bit) shows a flat floor rolling *off*. At 44 dB the passband matches 16-bit to 0.1 dB and the roll-off to 3 dB, with 16 dB of headroom above the strongest signal in the band. 50 dB buys 2 dB more in the roll-off for 6 dB of headroom, so 44 it is.

### Root cause

`DeviceInfo::digital_gain_db` added the HF+'s boost only `if self.resolution <= 8`. But `resolution` is the ADC width the server reports — 16 on this server, 18 in the crate's own fixture, never 8 for an HF+ — so the branch was dead and an 8-bit stream always went out at 0 dB. The doc comment on the function says "an Airspy HF+ *sending 8-bit* needs a further 32 dB"; the code checked the wrong thing, and the number was 12 dB short besides.

It was invisible before the first fix because the receiver was parked on the FM broadcast band, where there is plenty of signal for eight bits.

For the record, the comment attributed the formula to "the reference client". SDR++'s `computeDigitalGain` gives an HF+ decimation gain only (its `serverBits` argument is accepted and unused), so the boost is not from there — an HF+ over 8-bit in SDR++ starves the same way. The doc now says where the number comes from and carries the sweep.

### The fix

`digital_gain_db` takes the stream format and keys the HF+ boost on `SpyServerFormat::Uint8`, at 44 dB. Nothing else in the formula changes; a 16-bit stream from an HF+ has 48 dB more room and gets none of it, as before — and is the answer for a band with a station strong enough to clip eight bits at this gain, as is the manual digital-gain setting.

### Testing

- Same real HF+ over SpyServer: the sweep above.
- `an_hf_plus_gets_its_eight_bit_boost_whatever_resolution_the_server_claims` (unit): resolutions 8, 16 and 18 all get the boost as 8-bit and none as 16-bit.
- `an_hf_plus_sent_as_eight_bit_is_asked_for_its_forty_four_db` (integration, fake server describing itself exactly as the real one does, `resolution = 16`): `IQ_DIGITAL_GAIN` is 44 for an 8-bit stream and 0 for a 16-bit one.
- `cargo test -p sdroxide-spyserver`: 62 tests green.

### DIFF
--- a/crates/sdroxide-spyserver/src/net.rs
+++ b/crates/sdroxide-spyserver/src/net.rs
@@ -538,7 +538,7 @@ impl Client {
     /// whatever the operator pinned it to.
     fn digital_gain_wire(&self) -> u32 {
         let db = if self.auto_digital_gain {
-            self.info.digital_gain_db(self.gain_index, self.iq_stage)
+            self.info.digital_gain_db(self.gain_index, self.iq_stage, self.iq_format)
         } else {
             self.digital_gain_db
         };
--- a/crates/sdroxide-spyserver/src/net.rs
+++ b/crates/sdroxide-spyserver/src/net.rs
@@ -487,10 +487,13 @@ impl Client {
 
         self.set(proto::SETTING_IQ_FORMAT, proto::format_wire(self.iq_format))?;
         self.set(proto::SETTING_IQ_DECIMATION, self.iq_stage)?;
-        self.set(proto::SETTING_IQ_FREQUENCY, proto::freq_wire(self.center))?;
+        // The FFT window before the I/Q position, because the receiver follows
+        // the window and `IQ_FREQUENCY` is only a place inside the band the
+        // receiver is on — see [`flush_retune`].
         if self.fft_enabled {
             self.send_fft_config()?;
         }
+        self.set(proto::SETTING_IQ_FREQUENCY, proto::freq_wire(self.center))?;
         self.start_stream()
     }
 
@@ -584,8 +587,22 @@ impl Client {
         want.clamp(lo, hi.max(lo))
     }
 
-    /// The FFT window centred as near `hz` as its own slack allows.
+    /// Where the FFT window should be centred for the dial to be on `hz`.
+    ///
+    /// Where this client controls the receiver that is `hz` itself: with the
+    /// FFT lane running the receiver follows this window, so the window goes
+    /// where the dial goes. Clamping it into the slack around wherever the
+    /// server happened to be sitting is what parked whole sessions on the
+    /// server's own start-up frequency — at decimation stage 0 the window
+    /// covers the entire receiver and has no slack at all, so the clamp
+    /// collapsed to exactly that frequency and never let go.
+    ///
+    /// Where another client owns the receiver the device is theirs and does not
+    /// move, and all this end may do is slide the window inside it.
     fn clamped_fft_center(&self, hz: f64) -> f64 {
+        if self.can_control {
+            return hz;
+        }
         let slack = self.info.fft_slack_hz(self.fft_span);
         hz.clamp(self.device_center - slack, self.device_center + slack)
     }
@@ -598,7 +615,14 @@ impl Client {
         if !self.fft_enabled {
             return None;
         }
-        self.info.fft_recenter(self.center, self.fft_center, self.fft_span, self.device_center)
+        self.info.fft_recenter(
+            self.center,
+            self.fft_center,
+            self.fft_span,
+            self.device_center,
+            self.can_control,
+            self.info.iq_slack_hz(self.iq_rate),
+        )
     }
 
     /// Note an I/Q message's sequence number, reporting messages the server
@@ -734,7 +758,13 @@ fn apply(client: &mut Client, p: &Pending) -> Result<()> {
         // that changes what the server sends is worth.
         client.set(proto::SETTING_STREAMING_ENABLED, 0)?;
         if on {
+            // Where the window is *now*, not where it was left when the lane
+            // was last on: nothing has been maintaining it while it was off, so
+            // sending the stale centre would drag the receiver back to whatever
+            // band the operator was in then.
+            client.fft_center = client.clamped_fft_center(client.center);
             client.send_fft_config()?;
+            client.set(proto::SETTING_IQ_FREQUENCY, proto::freq_wire(client.center))?;
         }
         client.start_stream()?;
     }
@@ -772,20 +802,38 @@ fn flush_retune(client: &mut Client, shared: &Arc<Shared>) -> Result<()> {
     let sent = client.reachable_center(want);
     client.center = sent;
     shared.iq_center_milli_hz.store((sent * 1000.0) as i64, Ordering::Relaxed);
+
+    // The window moves first, and that ordering is load-bearing. On a server
+    // this client controls, the receiver follows the FFT window; `IQ_FREQUENCY`
+    // only places the I/Q window inside the band the receiver is already on, and
+    // a position outside it is clamped — and *stays* clamped, because the server
+    // holds it as an absolute frequency rather than as an offset. Placing the
+    // I/Q window before the receiver has moved therefore leaves it on the band
+    // being left behind.
+    move_fft_window(client)?;
     client.set(proto::SETTING_IQ_FREQUENCY, proto::freq_wire(sent))
 }
 
 /// Move the FFT window if the dial has reached its edge, or if the device
 /// moved out from under it.
 ///
-/// Runs every pass rather than only after a retune: on a server this client
-/// controls, tuning moves the *device* centre too, and the window's slack is
-/// measured against that — so a window that was in the middle of its range can
-/// find itself pinned to an edge without the dial having moved again.
+/// Runs every pass rather than only after a retune: on a shared server the
+/// owner can move the receiver with no dial move here at all, and the window's
+/// slack is measured against where they put it.
 fn maintain_fft(client: &mut Client) -> Result<()> {
     if client.last_fft_retune.elapsed() < RETUNE_MIN_INTERVAL {
         return Ok(());
     }
+    move_fft_window(client)
+}
+
+/// The move itself, without the rate limit.
+///
+/// [`flush_retune`] calls this directly: it has already paid the interval for
+/// the retune it is in the middle of, and on a server this client controls the
+/// window *is* the retune — a window held back by its own limiter would be a
+/// receiver held back with it.
+fn move_fft_window(client: &mut Client) -> Result<()> {
     let Some(target) = client.fft_target_center() else {
         return Ok(());
     };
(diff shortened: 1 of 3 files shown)


dividebysandwich:
That's *really* good, no complaints or comments!

PR #398, opened 2026-09-09
Implement nord and nord_dark color palettes

madmedicnl:
Added nord and nord_dark functions to define color palettes.


dividebysandwich:
Hi, thanks for your PR. Can you please extract the palettes into a separate single commit and make sure they're wired into the UI properly? I see all kinds of unrelated commits at the moment, including readme changes and 11m band support which conflicts with what's already in sdroxide.

PR #401, opened 2026-09-09
ui/theme: add the nord and nord_dark palettes

madmedicnl:
Two new palettes — **nord** (polar-night ground `#2E3440`, snow-storm ink, frost accents) and **nord_dark** (the same family scaled down toward black) — wired into the theme picker as new `UiTheme` variants.

- `crates/sdroxide-ui/src/theme.rs`: the two palettes as `const` entries in the 33-field `PALETTES` array (indexes 7 and 8); the S-meter reuses the classic navy-glass face (already the arctic look — cyan backlight, ice at S9); scope and map inks derive from each palette the way every dark theme's do.
- `crates/sdroxide-types/src/ui.rs`: `UiTheme::Nord` / `UiTheme::NordDark` added to the dropdown before the serde catch-all, with labels "Nord" / "Nord dark".
- The alert-colour test now asserts red *dominance* rather than one exact hue, so the authentic aurora red (`#BF616A`) passes while a TX/SWR indication still can never come out green.

Single commit, based on current `main`. `cargo test -p sdroxide-types -p sdroxide-ui -p sdroxide-config` and `cargo clippy -p sdroxide-ui` pass.

### DIFF
--- a/crates/sdroxide-types/src/ui.rs
+++ b/crates/sdroxide-types/src/ui.rs
@@ -297,6 +297,13 @@ pub enum UiTheme {
     /// faint. For low vision, for glare, and for a display that has lost its
     /// contrast.
     HighContrast,
+    /// Nord — the light arctic palette: polar-night ground and snow-storm
+    /// ink, frost accents. A dimmer, colder counterpart to the navy/cyan of
+    /// the default look.
+    Nord,
+    /// Nord Dark — the same arctic family on a darker ground: polar night
+    /// scaled down toward black, the frost and aurora accents unchanged.
+    NordDark,
     /// The classic navy/cyan/pink look. Declared last because serde demands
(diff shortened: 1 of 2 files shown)


dividebysandwich:
Ok I gave your theme a look and it's interesting! However there's a number of readability issues on Nord, for example on the decode lists, JS8 panel, help, etc... - The contrast of cq_bg is too low. You end up with white text on bright yellow.
Also, you loosen the every_theme_keeps_alerts_red test. I'd rather you choose a proper alert color.
The scroll_track is identical to row_bg, which makes the scroll bar gutter disappear, is that on purpose?

Nord Dark looks fine.

Please use c(0xRRGGBB) instead of Color32::from_rgb().

Would be great if you could address these, then we can merge!

madmedicnl:
committed changes and a few more common themes added. maintained your original colorcoding ways.

PR #402, opened 2026-09-09
Add AtCHAT NET digital protocol

barisdinc:
AtCHAT NET is a chat / file-transfer protocol built out of nostalgia for the old packet-radio days. It supports private and public chat and file transfer, and does all of it concurrently in an automatic-master mode — dynamic master election, a shared roster, and block-CRC-ARQ file/image transfers running alongside the keyboard traffic.

- Protocol overview: https://github.com/barisdinc/AtChat/blob/main/docs/OVERVIEW.pdf
- Protocol details: https://github.com/barisdinc/AtChat/blob/main/docs/PROTOCOL.pdf
- Standalone project: https://github.com/barisdinc/AtChat

This PR brings it into sdroxide as a digital protocol (`Mode::AtChat`), alongside the other digimodes.

## What's in it

- **`sdroxide-atchat` crate** — the OFDM modem (BPSK/QPSK, ~2.7 kHz USB), the NET protocol state machine (master election, roster ageing, common + directed chat, block-CRC-ARQ file/image transfer), and two channel backends:
  - the real radio (`RadioConnector`), with listen-before-transmit off the demodulated carrier (adaptive noise-floor gate, so a real receiver's AGC floor does not read as a permanent carrier);
  - a `channel_server`-compatible virtual TCP endpoint for radio-less development, opt-in via `DigiConfig::atchat_virtual`.
- **`AtChatController`** (`sdroxide-digi`) runs the whole protocol on its own thread and only moves 8 kHz PCM across the session rings, so nothing here can stall the engine's audio loop.
- **`Mode::AtChat`** in types + DSP, covered in the CAT / TCI / rigctld mode maps. New `Command::AtChat*` variants are appended after `SstvRestartRx` to keep postcard's positional numbering stable.
- **ATCHAT NET panel** in the UI — a roster strip, a tabbed conversation (a common CHAT tab plus a closable per-station DM tab), a LOG tab for on-air activity, and a FILES pane with the transfers in flight and a viewer for received images.
- **RF-less test tools** (`atchat-channel`, `atchat-client`, `atchat-monitor` binaries) and cross-implementation modem/protocol test vectors checked against the reference implementation.

Selecting the mode puts the station on the air; the virtual TCP channel is opt-in. Rebased onto current `main`; the only integration conflict was the appended `Command` / `Mode` variants.


dividebysandwich:
That's an excellent PR, fantastic work! Can't merge it in the current state, but there's not much missing. First of all, please do a cargo fmt pass (note our rulseset which is a bit more relaxed and produces compact code). Second, you enabled rt-multi-thread in tokio but never use it - consider removing it if not needed. Third and last big one: Documentation. This needs a chapter for the new AtCHAT mode in the user guide, ideally similar to the existing mode chapters. Oh and one minor one, in atchat_controller.rs you do ```out.len().saturating_sub(self.tx_fifo.len()).max(out.len())``` which is always ```out.len()``` - I don't think this was intended and I'm not sure if this won't cause some weird edge cases, could you take a look?

Really great addiition!

barisdinc:
Thanks for the review! All four addressed:

1. **cargo fmt** — done; the crate and its callers now match `rustfmt.toml`.
2. **Unused `rt-multi-thread`** — you were right, nothing here needs OS-level parallelism. `session.rs` already runs its own protocol/modem thread on a hand-built `current_thread` runtime; switched the three standalone bins' `#[tokio::main]` to `flavor = "current_thread"` too and swapped the tokio feature from `rt-multi-thread` to plain `rt`.
3. **Documentation** — added §3.17 AtCHAT NET to the user manual (master election/roster, the adaptive carrier-sense listen-before-transmit, joining/leaving, the virtual-channel loopback, the CHAT/DM/LOG panel, and file/image transfer including where a received file lands), cross-referenced from the ToC, the §1 feature overview, and the §16 mode table — the same three places every other mode is listed.
4. **The `saturating_sub(...).max(out.len())` bug** — confirmed and fixed. It always evaluated to `out.len()` (`saturating_sub` never exceeds it, so `.max` threw the subtraction away), meaning every top-up asked the session for a full block regardless of what the FIFO already had queued. It's now just the shortfall.

Four separate commits, one per point, on top of the existing branch.

hiperiondev:
AtCHAT NET github pages not exist!

barisdinc:
Hi, 
Somehow the repository was in PRIVATE mode, I changed that back to PUBLIC mode now.. its accessible with the history
Thanks for heads up
https://github.com/barisdinc/AtChat


Baris
TA7W/OH2UDS

PR #407, opened 2026-09-10
docs: QO-100 quick-start guide (EN + TR)

barisdinc:
Adds a task-oriented QO-100 quick-start that complements the reference material in `USER_MANUAL.md` §2.21 (QO-100 beacon plugin) and §6.2.7 (PlutoSDR).

**Files**
- `docs/qo100-quickstart.en.md` / `docs/qo100-quickstart.tr.md` — plain Markdown, ASCII panel sketches, no binary assets.
- `docs/qo100-quickstart.en.pdf` / `docs/qo100-quickstart.tr.pdf` — the same 19-step walkthrough rendered from an HTML source with drawn UI panel diagrams. A4.

**Contents** — 19 numbered steps in three parts:

- **Settings** — Radio tab: `PlutoSDR`, the `LNB, Ku low (-9750 MHz)` converter preset, `Transmit = Its own offset` at 0 Hz, Pluto address, `1 Msps`, `Full duplex`, Apply.
- **Locking on** — SAT window, pick QO-100, LOCK ON.
- **Beacon sync** — the QO-100 tab: `width`, one-shot `APPLY CORRECTION`, `ON` + `AUTO` for continuous LNB-drift correction, optional `TELEMETRY`.

The Markdown panels use ASCII sketches in the same style as the existing §2.21 diagram; the PDF carries the same steps with drawn diagrams. The Turkish files are a straight translation with the same step numbers.

Happy to drop the PDFs and keep Markdown only, or to add a pointer from §2.21 — whichever you prefer.


dividebysandwich:
Thanks, that's pretty awesome! Some minor nitpicks:

1. this guide isn't linked to from anywhere. A prominent link from readme and the manual would be good.
2. Can you add the html sources for the PDFs, so the PDF versions of the guide can be amended later? Or commit the images and use those in the markdown?

barisdinc:
Thanks for the review! Both addressed:

1. **Linked** — the guide is now linked from the README (right under the User Manual link) and from `USER_MANUAL.md` §2.21 (QO-100 beacon plugin), which links back to it too.
2. **PDF sources** — replaced the ASCII UI sketches with drawn panel mock-ups (styled from the app's own palette in `theme.rs` and its Chakra Petch / Share Tech Mono fonts) instead of committing a separate HTML source. They're PNGs under `docs/images/qo100-quickstart/`, embedded with plain Markdown `` syntax so they also render on GitHub, and `tools/gen_qo100_quickstart_pdf.py` (re)draws them and rebuilds both PDFs straight from the two Markdown files. So there's still just one source of truth — editing a step and its diagram is a one-file change, and the PDFs regenerate from it.

PR #450, opened 2026-09-15
Decodes: export the decode list as CSV and received-report ADIF

### LINKED ISSUE #433: Feature Request - Logging / export (FT8,...) for short wave listeners
Please let us non licensed short wave listeners export/log received FT8, FT4,... decodes. 

Currently the only way seems to download uploaded reports to PSKreporter, would be great to have a local functionality, as we have no QSOs and that seems to be neccessary for the included export funktionality.

THANK YOU for your amazing project!

madmedicnl:
Decodes: export the decode list as CSV and received-report ADIF

A short-wave listener has no contacts, so the logbook's ADIF and TXT exports
had nothing to write and the only way out of the decode list was PSK Reporter
(fixes #433). Add digi_decodes_to_csv and digi_decodes_to_adif: one row per
decode keyed by the receiver's dial plus the decode's audio offset, with the
member band name in the CSV (ADIF cannot enumerate 11 m) and, in the ADIF,
CALL set to the station heard and no RST_SENT/RST_RCVD — a received report
that does not pretend to be a QSO. Two buttons in the DECODES header.


Promethium147:
Thank you so much for your work, appreciate that A LOT!

Issue #322, opened 2026-09-04
Feature Request: PC-keyboard as a Morse key?


fpausp:
I would like to use a key on my PC keyboard as a Morse key to make CW QSOs. Perhaps it would even be possible to install a 3.5 mm jack and connect it to that particular key used for CW. Then you could plug a real Morse key into the jack and use it that way.

PR #453, opened 2026-09-15
CW: the PC keyboard as a straight key

madmedicnl:
CW: the PC keyboard as a straight key (#322)

The operator holds Space as a key: down while held, up on release. The whole
keyer is handed to the key and timed text drops — nothing from a half-typed
automatic message surfaces between elements. An SDR drives the sidetone
through its transmit chain; a rig that keys itself from text has nothing a
hand key can drive, so the controller refuses to engage there. The transmit
box is locked while the mode is on, and a PTT bound to Space is swallowed so
it does not fire under a keyed hand. A focused widget steals the key: a caret
in another field means a typist, not a keyer. The button lives in the
transmit row beside TX, and the mode resets on a mode change so a stale
toggle does not surprise a later operator.


dividebysandwich:
Issues found: 

Holding Space while switching radio tabs leaves the carrier on until you come back. There's no maximum key-down time.

The key is only sampled once per 50 ms chunk (400 samples at 8 kHz), which ruins element timing.

CwStraight/CwKey are inserted mid-enum with no protocol bump.

madmedicnl:
Thanks — all four addressed, and rebased onto current `main`:

- **Carrier left on across a tab switch.** The key is now let go of in `release_held_controls`, so losing the keyboard (tab switch, window blur) drops it. The *mode* stays engaged; only the key goes up.
- **No maximum key-down.** A continuous hold is capped at 30 s; the over ends and the status sets `tx_watchdog`, which the CW panel now shows as a **WATCHDOG** chip. Keying again clears it.
- **50 ms sampling.** The straight-key path no longer goes through the 8 kHz `TX_CHUNK` (400 samples = 50 ms) and its resampler. It renders directly at the output rate and reads the key once per sample, so the resolution is the engine's own block. A test presses for 15 ms and asserts the keyed span stays near 15 ms, not 50.
- **Protocol.** `CwStraight`/`CwKey` moved to the end of `Command` and `PROTO_VERSION` is 149.

Added two tests for the DSP side (short element, hold cap). Not tested here: a real rig, and the web client.

Issue #197, opened 2026-08-27
SDROXIDE Profiles request


totally1liquid-source:
It would be great to be able to create profiles within SDROxide. I'm constantly having to adjust gains etc when moving around bands. I'm running a SDRPlay RSPDXR2. SDRUno allows you to create mem profiles and also settings profiles. Loving the app BTW. Antenna selection would also be great to have on the main interface. This SDR has 3 inputs.

soakland5:
I second this. We operate a club station at a primary op loction, and it would be nice to be able to seamlessly switch between profiles. 73 de Scott, KG4O

PR #454, opened 2026-09-15
Profiles: save and re-apply how the station is set up

madmedicnl:
Profiles: save and re-apply how the station is set up (#197)

A profile is a named snapshot the operator takes on purpose — contest, DX,
CB, the band plan. It captures the whole rememberable radio state (dials and
VFOs, mode and filters, gains, drive and antennas), the digital identity and
message templates, and the band stacks, and applying it puts all of that
back in one click.

The hardware is deliberately left out: the back end, the audio devices and
the converters are personal to a radio, not to a way of working it, so a
profile follows the operator across the rig they always sit at rather than
dragging yesterday's sound card behind it. An apply works through the same
paths a band change or a session restore do — the active VFO retunes whole —
so nothing is reopened and the front end is never disturbed.

Ships as a Profiles tab in settings: a name box and Save beside a list of
saved setups with Apply and Delete. The engine owns the store (profiles.json,
station scope) and answers every action with the fresh name list, so the
screen never holds the data, only the offer. Saved through the session's
own snapshot logic, refactored out of save_session so a profile captures
exactly what the operator last left the radio in.

---

Rebased onto upstream main with the #322 straight-key commit as its base, since the session work it refactors was landed alongside it. One merge conflict resolved during the port (settings tab order). No test changes upstream.


dividebysandwich:
Some issues here:

A profile saved on VFO B applies VFO A's frequency.

On a multi-radio station each radio rewrites profiles.json from its own copy, so profiles get lost.

The UI's copy of the digital settings isn't refreshed, so the next edit puts the old callsign back.

Apply isn't refused while transmitting, writes session.json even on engines that must not save a session, and has no protocol bump.

madmedicnl:
Thanks — all four fixed, rebased onto current `main`:

- **VFO B applies A's frequency.** `Session::freq_hz` is VFO A's dial whatever is active, and the apply read it as the active dial. It now uses `vfo_b_hz` when B was the one in use. Test: save on B, move to A, apply, assert B's dial comes back.
- **Multi-radio lost profiles.** Each radio rewrote `profiles.json` from its own copy. `reload_shared_stores` now reloads them, and a save/delete catches up with the shared store first and announces its own write after — the same pattern memories and band stacks use.
- **Stale digital copy in the UI.** Applying a profile now marks the screen's editable `DigiConfig` for re-seeding from the next status, so the next edit can't put the old callsign back.
- **Apply guards.** Refused while transmitting, and the post-apply `session.json` write now only happens on an engine that remembers a session.

`PROTO_VERSION` is 149 with the three commands appended last (the rebase fixed the missing bump).

One thing I noticed while testing: a `profiles.json` written by an older build (before upstream's `Session` changes) fails to parse and is quarantined, so the profile is lost on upgrade. Not in your list and I left it alone — but if you want a migration shim for the embedded `Session`, say so and I'll add one.

PR #457, opened 2026-09-15
UI: audible alerts for calls, directed CQs and new DXCC/grids

madmedicnl:
UI: audible alerts — calls, directed CQs and new DXCC/grids ring on their own audio output

A Settings → Alerts tab, the other half of the spoken-announcements story on
the UI tab: speech reads changes out, these ring when a decode matters, and no
radio path is touched — the alarm plays over its own output stream, so the band
in the headphones and the speaker in the room can carry the same session
separately.

Five events, each with its own switch and its own sound (synthesised, not
shipped as files), mirroring the decode list's badges: a station calling us
(to, or rr73_to in Hound mode), a directed CQ we could answer (off by default —
a busy evening is a hundred a minute), a new DXCC entity, a new entity on this
band, and a new grid square. The matcher is pure, in sdroxide-types, so the
lights and the sounds can never disagree; a per-station cooldown keeps a
station that decodes every slot from ringing every slot.

The alarm is deliberately not gated on window focus: it exists to reach the
operator when they are looking at another window entirely. The hook only reads
the log's novelty set, and is skipped while alerts are disabled.

Native only, like speech's synthesizer: a worker (same ring-pacing discipline
as the speech engine) renders the tones into sdroxide-audio's own output. The
browser build keeps the settings for the sink that has not landed there. The
table is fields-first like [speech], and sdroxide-config guards that ordering
with its usual read-back test.


dividebysandwich:
That's a really nice feature addition! However there's a few issues with this PR:

The app freezes on quit or when alerts are turned off if the alert device failed to open. The worker drains its channel waiting for a close that never comes, because Drop still holds the sender while it waits for the worker.

Instant::now() isn't gated on wasm, so the web client crashes on the first alert.

Changing the output device does nothing, and one decode batch can queue something like 16 alerts.

madmedicnl:
Thanks — all three fixed, rebased onto current `main`:

- **Freeze on quit / alerts off.** The failed-open worker waits for its channel to close, and `Drop` held the sender across the join, so it never closed. `AlertSink` now keeps the sender in an `Option` and drops it *before* joining, and the worker honours an explicit `Quit` too.
- **wasm crash.** The cooldown was on `Instant`, which panics on `wasm32-unknown-unknown`. It counts wall-clock seconds through `crate::time::now_unix_f64` now; these are minute-long cooldowns, so a stepping clock is fine.
- **Device change did nothing.** `sync_sink` only starts a worker when there isn't one, and there already was. A device change now drops the old worker first, then reconciles — so the new device is actually opened. And one batch raises **one** alarm (the first match) instead of one per matching decode; a test pins that.

Not tested here: a browser build with alerts enabled, and a second physical audio device.

PR #465, opened 2026-09-16
ACARS: the airband airline datalink, as a receive mode

### LINKED ISSUE #436: ACARS Support
Support for ACARS around 130MHZ beyond VDL2 support already added.

madmedicnl:
Closes #436.

ACARS is the VHF aircraft datalink around 130 MHz: an AM carrier in the airband carrying 2400-baud MSK on 1800 Hz, NRZI-coded, character-oriented with odd parity and a CCITT block check.

**Decoder core** (`sdroxide-dsp/src/acars.rs`)
- `parse_frame` turns an NRZI-decoded bit stream into a message, hunting for SOH and keeping whichever heading ends in a matching block check.
- `AcarsRx` mixes 1800 Hz down, low-passes, discriminates the phase, integrates a symbol and undoes the NRZI.

**Mode and panel**
- `Mode::Acars`, receive-only, grouped with AM in every CAT backend and speech label.
- A live panel with the published airline channel chips, the message list (address, mode, label, text, time), and the signal level between bursts.
- The status rides the existing digi status/config channel, so no protocol bump.

**Testing:** the decoder is unit-tested against a synthetic frame, a corrupted one (the block check must catch it), noise (no phantom) and the CRC's known answer, plus an end-to-end synthetic signal. `cargo check --release --workspace` and the `sdroxide-dsp`/`sdroxide-digi`/`sdroxide-types` tests pass.

**Not verified on a real off-air signal** — the demod samples on a fixed grid rather than recovering timing, so someone with an airband antenna should confirm before this is trusted on the air. Flagging that rather than claiming otherwise.


dividebysandwich:
I am not sure this will decode actual signals.

The checksum is CCITT-FALSE (0x1021, init 0xFFFF, big-endian). The repo's own VDL2 ACARS parser uses Kermit (0x8408, init 0).

The symbol clock drops the fractional sample on every symbol, so timing drifts at common rates such as 51.2 kHz.

The NRZI bit decoding probably runs the wrong way, compared with acarsdec's source.

A field is inserted into the middle of DigiStatus with no protocol bump, which breaks every digital mode for mismatched remote clients.

CAT rigs get set to USB/DATA instead of AM.

The tests use the decoder's own encoder, so they can't catch any of this. A recording would clear this up.

madmedicnl:
Thanks — worked through these, and rebased onto current `main`.

**Fixed:**

- **Symbol clock.** It cleared the sample counter every symbol instead of carrying the fraction, so at 51.2 kHz (21.33… samples/symbol) the grid walked off the signal across a frame. It now carries the overshoot, and there's a test that decodes a frame at 51.2 kHz — it fails on the old code and passes now.
- **`DigiStatus` field.** Moved `acars` to the end of the struct (with `serde(default)`) and bumped `PROTO_VERSION` to 149. It was mid-struct, which shifts the postcard tail for a mismatched peer.

**On the others, what I found:**

- **Checksum** is CCITT-FALSE — `crc16` is 0x1021/init 0xFFFF, big-endian, and there's already a known-answer test (`"123456789"` → `0x29B1`). The Kermit one you saw is the VDL2 parser's, separate. If you're seeing real frames fail the check with a correct CRC, that points at the bit timing rather than the polynomial.
- **CAT mode** is AM in every map I can find: CI-V `0x02`, Kenwood/Yaesu/ELAD/Elecraft/QRP Labs `'5'`, flrig `"AM"`, rigctld `"AM"`. If you saw USB/DATA it may have been an earlier revision; if it reproduces on current `main` of the branch I'd like to know which rig and which protocol.
- **NRZI** is change→0, no-change→1, which matches the spec's convention; I compared against acarsdec's `msk.c`, where the phase detector folds the NRZI into `putbit`'s sign, so it isn't a like-for-like diff. I don't want to flip it blind.

**The recording is the real gap.** The tests still drive the decoder with its own encoder for the full path, so they can't tell whether the demodulator matches the air — the fractional-rate test now catches the timing class of bug, but not an inverted NRZI or a wrong level convention. If you can drop in a short off-air recording (audio or IQ, any rate) I'll add it as an ignored fixture test with real assertions, the way the DRM/HD ones do. Otherwise I'd rather leave the NRZI as-is than guess.

madmedicnl:
I found a reference recording (acarsdec's own `test.wav`) and can reproduce the problem: the decoder produces no frame from real audio. It has no carrier or bit-clock recovery, so it only decodes its own encoder's output. I'm reworking the demodulator against that recording now and will report back with the result rather than going back and forth here.

madmedicnl:
Status: confirmed against a reference recording, and the fix is bigger than the checksum.

I built acarsdec itself and ran its `test.wav` through it — it decodes 7 messages across the four channels. Our decoder produces **zero frames** from the same file, so the "won't decode real signals" concern is real, not theoretical.

I then ported acarsdec's `msk.c` demodulator (half-sine matched filter, bit clock, PI carrier loop) and compared bit-for-bit against its output. What I get:

- fixed-grid integrator (what we have): no frame at all
- matched filter, loop always on: ~67% of bits match acarsdec's
- + signal-power gate: worse (~18%)
- + its 0xFF PREKEY gate: ~68%
- + its polarity-inversion handling: ~71%

70% of bits is nowhere near enough for the block check, so a partial port isn't worth shipping. The missing half is `acars.c`'s byte-level state machine (`decodeAcars`: PREKEY → SYNC → header → text → CRC, with the per-state `nbits` shifts and the polarity flip). It isn't just the parser — it's what drives the demodulator's PLL gating, so the two have to be ported together. That's the actual reason the decoder only ever worked on its own encoder.

Porting the demod + state machine together against `test.wav` is the next step; I'll report back with the result. I'd rather not push a guess in the meantime.

madmedicnl:
Done — and you were right on both counts: it would not decode real signals, and the tests using the decoder's own encoder are what hid it.

Checked against acarsdec's off-air `test.wav` (decodes there, did not here), two things were wrong:

**The block check was not CCITT-FALSE.** I believed your bullet and went looking for a mismatch elsewhere, but the bytes disagree: ACARS is a **reflected** CRC-16 (polynomial 0x8408, init 0) over the bytes **as received, parity bits included**, from the mode character through ETX, and the two BCS bytes carry the **low byte first**. I verified this on two frames from the recording (e.g. `…2f 83`, BCS `61 e5`): the check runs to zero under that rule, and does not under CCITT-FALSE over parity-stripped bytes with SOH prepended. The VDL2 parser's Kermit is the same reflected polynomial; the difference is the init value and the parity.

**There was no carrier or bit-clock recovery.** The fixed grid only worked on our own encoder. The demod is now acarsdec's `msk.c` — half-sine matched filter, bit clock off the VCO, PI carrier loop — with the input resampled to the 12 kHz those constants are defined at (the way acarsdec averages its input down). Re-deriving the constants per rate does not work; 48 kHz produced nothing, which matters because the engine feeds the decoder at the audio rate.

Result: `an_off_air_recording_decodes` (ignored, `SDROXIDE_ACARS_SAMPLE`) decodes the real `F-GTAE H1` frame at both 12 kHz and 48 kHz. The two synthetic end-to-end tests are removed — acarsdec cannot decode our synthetic audio either, so they were testing a signal that does not resemble ACARS.

Not done: acarsdec's error *correction* (its syndrome search fixes a few parity/CRC errors). This decodes frames that are clean, which the recording's happen to be; a weak channel that needs the FEC will still fail the check rather than be corrected. Worth deciding whether you want that ported too.

madmedicnl:
Rebased onto current `main`. `PROTO_VERSION` is now **154** — your 153 plus a v154 entry for ACARS' trailing `DigiStatus` field — so it's ready to merge without a numbering conflict. Fixture still decodes a real frame at 48 kHz.

PR #466, opened 2026-09-16
HD Radio (NRSC-5): the FM digital sidecar as a receive mode

### LINKED ISSUE #437: NRSC5 Support
HD Radio for North America.

madmedicnl:
Closes #437.

HD Radio is the digital sidecar of an FM broadcast: OFDM carriers about 20 dB below the analog carrier, carrying HDC audio plus the station's name, slogan and message. This adds it as a broadcast mode, decoded by the vendored [nrsc5](https://github.com/theori-io/nrsc5) receiver.

**Vendored decoder**
- `vendor/nrsc5` added as a submodule (`theori-io/nrsc5 @ 0225922`, GPL-3.0-or-later).
- `vendor/faad2` re-pinned to `madmedicnl/faad2-hdc` — stock 2.11.2 plus nrsc5's `support/faad2-hdc-support.patch` — and built with `HDC_SUPPORT` alongside `DRM_SUPPORT`, so **one** faad2 archive serves both the Dream DRM and the nrsc5 decoders. Two copies in one binary collide on every `NeAACDec*` symbol.
- New `sdroxide-nrsc5` crate builds the vendored library with `cc` (bypassing nrsc5's CMake), supplies a radix-2 `fftwf` stand-in for the power-of-two transforms, and weak `librtlsdr` stubs for the device path the pipe API never reaches. It exposes `HdReceiver`/`Mode`/`Event` over nrsc5's pipe.

> **On the faad2 pin:** this is the one part that needs a maintainer decision. The alternative is merging nrsc5's HDC patch into `knik0/faad2` first, after which the pin can go back to upstream. I've kept everything else independent of it.

**Mode, engine, panel**
- `Mode::HdRadio`, receive-only, grouped with FM in every CAT backend (no CAT table carries a frequency to tell FM from AM HD).
- `HdDemod` resamples the channel to the FM hybrid's 744,187.5 S/s, normalises the level, pipes I/Q to the decoder and meters the decoded audio out.
- Engine poll publishing `HdRadioStatus`, a retune reset, and `Command::SetHdProgram`.
- Server/protocol at v147 (`ServerMsg::Hd`, cached and replayed on connect), plus a station-information window and an HD top-bar chip (sync, both sidebands' MER, CBER, offset, PSMI, name/slogan/message, HD-2/HD-3 chips).

**Testing**
- `cargo check --release --workspace` clean; the `sdroxide-types`, `sdroxide-nrsc5`, `sdroxide-proto` and `sdroxide-dsp` suites pass.
- The `fftwf` stand-in is verified against a naive DFT; the vendored `sample.xz` decodes end to end (2 syncs, 492 audio frames, "KUT / The University of Texas at Austin").
- **Not exercised:** live I/Q through the engine. The demod samples on a fixed grid and I had no FM capture to run it against, so this should be treated as needing an off-air check before it is trusted.

**Not in this PR:** HD on the AM band (the decoder wants opening in AM mode and feeding at 46,511.71875 S/s, chosen from the dial), album art/PSD, and receiver blending back to analog when the digital side drops.

**Credit and prior art**

The design this follows was worked out and measured in #437 by @dielectric-coder before these commits existed, and is owed to them:

- **one** faad2 archive carrying `DRM_SUPPORT` + `HDC_SUPPORT`, because two copies collide on every `NeAACDec*` symbol;
- the rates — 744,187.5 S/s for the FM channel, and 46,511.71875 S/s for the AM variant, which sdroxide's own decimation already lands on;
- **MER per sideband** rather than as a single number, which came from measuring a station whose sidebands arrived about 8 dB apart (CBF on 95.1 MHz).

They validated the combined faad2 against a real DRM broadcast (Radio Romania, 13.750 MHz, 130.8 s decoded) and across 20,000 frames of pseudo-random parametric-stereo input, byte-identical against stock. That matters here because `HDC_SUPPORT` is **not inert for DRM**: the patch switches faad2's own DRM paths on and alters its parametric-stereo code. They have also offered to run this PR against four local HD stations on an RSPdx (sync, per-sideband MER, CBER, audio on each programme, and behaviour as the signal degrades) and to repeat the DRM regression; that testing is pending.


dielectric-coder:
Tested on air, on an RSPdx against the four HD stations that reach me. Branch `upstream-pr/437-hd-radio` at `b012c692`, built on Arch with the submodules as pinned. Thanks for the credit in the description.

**It locks and decodes on every station I can hear.** Station identity, service lists and programme selection are all correct — HD1/HD2/HD3 on 107.3 are three genuinely different streams.

| Station | MER lower / upper | CBER | Services | PSMI |
| --- | --- | --- | --- | --- |
| 95.1 CBF | 13.6 / 5.9 dB | 0.0083 | 2 | 1 |
| 99.5 CJPX-HD | 12.4 / 12.9 dB | 0 | 2 | 1 |
| 107.3 CITE-FM | 12.2 / 12.1 dB | 0 | 3 | 3 |
| 107.9 WVPS | 13.6 / 13.7 dB | 0 | 3 | 3 |

CBF is the interesting one: its lower sideband arrives about 8 dB stronger than its upper and it still decodes at 0.8% CBER. Showing both MER figures rather than one is doing real work there.

**The faad2 pin is safe for DRM.** `madmedicnl/faad2-hdc` at `733c2ee` is stock 2.11.2 plus one commit applying nrsc5's patch, and its tree is byte-identical to the one I tested in the spike. Re-running the DRM regression against your pin: a real broadcast (Radio Romania "SNR Tiganesti E1" on 13.750 MHz, mode B, 16-QAM, AAC) decodes **byte-identically** to the stock `vendor/faad2` build — 130.8 s of audio, same bytes. With the 20,000-frame parametric-stereo comparison from the spike, that half of the change looks clear to me.

## One real bug: the decoder's audio is consumed as mono, but nrsc5 delivers interleaved stereo

`HdDemod::process` queues the raw `i16` values from `Event::Audio`, then computes `want` as a count of audio *frames* and pushes one queued value per frame. nrsc5 supplies two values per frame at 44,100 Hz, so half the audio is never consumed and what is emitted is the interleave itself.

Measured on a 23.66 s capture of 95.1 MHz: **1,044,146 values out, i.e. 44,133 per second** — exactly one per mono frame period, against the 88,200 the decoder produced.

Three consequences, all audible:

- playback runs at **half speed**, with L and R alternating;
- the queue grows without bound, so the backlog trim fires constantly — **55 times in those 23.66 s**, dropping ~15,000 values each, roughly every 0.4 s;
- **stereo is lost**.

`MAX_BACKLOG_FRAMES` has the same units mix: `self.audio.len()` counts values, the constant counts frames.

Reproduced two independent ways, so it is not my test rig:

1. driving `HdDemod` directly from a capture at the native rate (which skips your fractional resampler entirely);
2. listening in the GUI on this branch, RSPdx at 2 Msps through the normal chain — the operator's description was "half speed and stuttering" before being told what to expect.

`DrmDemod::process` is the model to follow: it pops in pairs and pushes `(l + r) * 0.5` to `out` with `(l - r) * 0.5` to `side` for the stereo blend. The same shape here fixes the rate, the backlog and the stereo together.

## Notes on your two points

- **The resampler.** Understood, and my harness fed the native rate deliberately, so the fractional path was only exercised by the GUI test above. Nothing there looked wrong once the pacing bug is accounted for.
- **Silence when the digital side drops.** Taken as expected behaviour; nothing in my figures is a lock that failed to produce audio. Every station above locked and played.

Happy to contribute the capture harness as an example in the crate, and to send a capture so the pacing can be checked without an antenna. I'll re-test on air once the audio path is fixed.

dielectric-coder:
A second finding, from trying to reach the mode by hand rather than through a harness: **HD Radio cannot be selected in the interface.**

The mode menu builds its rows from explicit lists — `Mode::Drm` is in the analog row in `top_bar.rs`, and `Mode::HdRadio` is in neither that row nor the digital one below it. Everything downstream is wired: `Mode::ALL`, the engine, `RxChip::Hd`, the panel. There is simply no chip to click. The only ways in are a memory channel (the memory editor iterates `Mode::ALL`) or hand-editing `mode` in `session.json`.

That is presumably also why the audio fault went unnoticed: with no way to select the mode from the UI, nobody listened to it for long.

And on listening — reached through a memory channel, on an RSPdx at 2 Msps, the operator's unprompted description of the decoded audio was "half speed and stuttering". Which is what the pacing bug above sounds like from the chair: everything an octave down, sloooowed out, with a skip about twice a second as the backlog is trimmed. Station name, slogan and the programme chips are all correct while that is happening, so the decode itself is clearly fine — it is only the audio path out of it.

Both of these are small: one line in the mode row, and the pairwise pop in `process` that `DrmDemod` already shows.

madmedicnl:
Thank you — this is exactly what the "not tested on air" caveat was waiting for, and both findings are right.

**The audio pacing is fixed** (pushed as `45bf39e0` on the branch, so `b89bff68` at the tip). A frame is now taken as a stereo pair, exactly as `DrmDemod` does:

- `(L + R) / 2` goes to the mono output the receive chain plays;
- `(L - R) / 2` goes to a side buffer offered through `take_side`, so the stereo blend works;
- `MAX_BACKLOG_FRAMES` / `TARGET_BACKLOG_FRAMES` are now counted in frames against `audio.len() / 2`, which is what the constants always meant.

The pairing is a free function with its own tests (a frame consumes two values and yields the sum and difference; a short queue is padded rather than shortened), so that one cannot regress without a decoder. I have **not** re-run it on air here — no FM front end — so your re-test is the evidence that matters.

**The missing menu chip is fixed too** (`c0119aa7`): `Mode::HdRadio` now sits in the analog mode row beside `Mode::Drm`, which is where a broadcast demodulator belongs rather than in the digi-engine row. That is also the reason nobody heard the pacing bug for long, as you said.

On the rest:

- **The faad2 pin** — thank you for re-running the DRM regression against `733c2ee` and confirming the real-broadcast audio is byte-identical. That was the part of this change most likely to bite someone, and it is now covered by your measurement rather than only by the fork's 15/15 DRM suite.
- **The capture harness as an example** — yes please. If you are willing to **send a capture**, that is better still: it would let the pacing be checked in the crate's own test suite without an antenna, on every change, which is the one thing this PR has lacked from the start.
- **MER per sideband** already had its case made by 95.1 at 13.6 / 5.9 dB — nice to see it confirmed against a second set of stations.

Please do re-test when you have the chance, and thank you for the thoroughness on this.

dielectric-coder:
Re-tested on air at `b89bff68`, same RSPdx, same four stations, same captures as before. **Both fixes confirmed.**

| Check, 95.1 MHz over 23.7 s | before | after |
| --- | --- | --- |
| backlog drops | 55 | **0** |
| values emitted | 44,133/s — one per mono period, half what the decoder produced | correct |
| spectral centroid vs an independent decode of the same capture (637 Hz) | 529 Hz | **640 Hz** |
| envelope correlation with that decode | 0.42 | **0.80** |

All four stations still lock with the same figures and identities, and none drops audio now:

| Station | MER lower / upper | CBER | Services | Backlog drops |
| --- | --- | --- | --- | --- |
| 95.1 CBF | 13.6 / 5.9 dB | 0.0083 | 2 | 0 |
| 99.5 CJPX-HD | 12.4 / 12.9 dB | 0 | 2 | 0 |
| 107.3 CITE-FM | 12.2 / 12.1 dB | 0 | 3 | 0 |
| 107.9 WVPS | 13.6 / 13.7 dB | 0 | 3 | 0 |

**The stereo side channel carries real signal**, and the levels are the right shape for the content: RMS 0.055 on 107.3 (music) against 0.0093 on 95.1 (Radio-Canada Première, mostly speech).

**The menu chip works.** HD Radio is selectable from the analog row now, no memory-channel trick needed. Listening in the GUI on 107.3 and 107.9: the half-speed and the twice-a-second stutter are both gone, and the programme chips switch between three genuinely different streams.

## On the capture

Happy to send one, with one caveat worth deciding first: a capture is somebody's copyrighted programme material, which is the reason `sdroxide-drm` keeps DRM recordings out of the tree and reads one from `SDROXIDE_DRM_SAMPLE` instead, skipping the test when it is not set. Size points the same way — 20 s at the decoder's own rate is ~140 MB, and even a 3 s clip is ~21 MB.

So my suggestion is to follow the DRM precedent rather than commit audio: an `#[ignore]`-free test that reads `SDROXIDE_HD_SAMPLE`, decodes it and asserts lock, station name, per-programme audio and **no backlog drops** — that last one is what would have caught this. I'll send you a capture for your own bench either way; say where you want it.

I'll also contribute the harness as `examples/hd_capture.rs`: it takes an `--record-iq` capture, shifts one channel to zero, decimates and drives `HdDemod` directly, printing lock, MER, CBER, identity, side-channel RMS and the drop count. That is what produced every number above, and it needs no antenna once a capture exists.

madmedicnl:
That is the confirmation this needed — thank you. The three numbers together are what makes it convincing: **0 backlog drops**, a spectral centroid of 640 Hz against the independent decode's 637, and envelope correlation up from 0.42 to 0.80. And the side channel carrying 0.055 RMS on the music station against 0.0093 on the speech one is the stereo path doing exactly what it should, rather than a plausible-looking zero.

**On the capture — agreed, let us follow the DRM precedent and not commit programme material.** Please do **not** put a recording in the tree. What I would like is:

- you send a short capture (a link you control — object store, your own site, or an asset on your fork — is fine; a 3 s clip at the decoder's rate is ~21 MB and plenty), and
- the branch gets a test keyed on `SDROXIDE_HD_SAMPLE`, which prints a line and passes when the variable is unset, exactly as `sdroxide-drm`'s `a_recording_decodes` does with `SDROXIDE_DRM_SAMPLE`.

The assertions you suggest are the right ones: lock, station name, **audio on each programme**, and **no backlog drops**. That last one needs the demod to expose it — there is no count today — so I will add a `backlog_drops()` counter to `HdDemod` and assert it is zero. You are right that it is the assertion that would have caught this; better that it fails loudly next time than that someone has to listen for half speed.

**The harness as `examples/hd_capture.rs` — yes please, that is the right home.** It needs no antenna once a capture exists, which makes it useful to anyone touching the demod. Open it against `madmedicnl:upstream-pr/437-hd-radio`, or post it and I will take it in.

I will hold off writing the test until your capture lands, so the conversion from a `--record-iq` capture to channel samples lives in one place — your harness — rather than being duplicated in the test.

madmedicnl:
The bench harness landed on this branch: `crates/sdroxide-nrsc5/examples/hd_capture.rs` (@dielectric-coder contributed it against the fork's `upstream-pr/437-hd-radio`, and it is the one that produced the on-air numbers above).

```
cargo run --release -p sdroxide-nrsc5 --example hd_capture -- cap.cf32 <centre Hz> <channel Hz> [rate] [programme]
```

It reads an `sdroxide --record-iq` capture (or a clip already cut to one channel at 744,187.5 S/s, CF32 or interleaved `.cs16`), shifts the wanted channel to zero, decimates and drives `HdDemod` directly — no antenna, no GUI — and prints lock, per-sideband MER, CBER, carrier offset, PSMI, station name/slogan, the service list, the stereo side channel's RMS, how much audio came out, and the backlog drop count (zero on a healthy stream).

A short capture is on its way for a test gated on `SDROXIDE_HD_SAMPLE`, which will assert lock, station name, audio on each programme and `backlog_drops() == 0` — the assertion that would have caught the pacing bug — skipping with a printed line when the variable is unset, as `sdroxide-drm`'s does.

dividebysandwich:
Hey, I've already started merging this PR. Please put any improvements into a new PR.

If you're still working on it, please put your PRs into "draft" mode.

madmedicnl:
Closing this as **superseded**: HD Radio is now on `main` (the decoder crate, a `sdroxide-faad2` that applies nrsc5's patch at build time, the mode, panel, protocol, and the follow-up fixes), so this branch has become a duplicate — which is why it now shows as conflicting.

I diffed the two. Everything this branch carried is in `main` in some form, including `backlog_drops()`, the stereo pair handling in `process`, and the mode-menu chip. The one thing `main` does **not** have is @dielectric-coder's bench harness, so that is now its own small PR — **#479** — adapted to the crate as it stands (the `sdroxide-faad2` dependency and the `HdDemod` API on `main`), carrying their authorship and the reasoning behind it. Nothing else of theirs is lost.

Thanks for the review and the on-air testing that got this in.

madmedicnl:
Understood — and thank you for taking it in. I see `main` now carries it (with the example and your `f3c4712f` fix on top), so I have closed the duplicate example PR (#479) rather than leave it in the way.

Two things this reorders on my side: the improvements I had queued go up as new PRs against `main` (the open ones from the fork — #476 web microphone, #477 config robustness, #478 JS8 composer, #454 profiles, #465 ACARS — are finished changes rather than works in progress), and anything still half-done will be a draft from now on.

PR #476, opened 2026-09-17
Web: connect the mic worklet to the destination so it is rendered

### LINKED ISSUE #468: Litlle bug with Web client: microphone
**Title:** Web client: microphone initializes (`mic ready`) but no audio frames are ever sent to the server during PTT — voice TX silently transmits nothing

## Environment
- SDRoxide version: 1.6.7
- Backend: CAT/Audio (Zastone ZT-7500, "Wolf" firmware, Kenwood TS-2000 emulation), paired with a HackRF Pro as a borrowed panadapter receiver
- Server: Debian 13 (trixie), Intel N100, Proxmox VM, reached via Caddy reverse proxy (`https://sdrxxxxxxx.com/`, valid Let's Encrypt cert)
- Clients tested: **web client only** — Chrome on Windows 11 (desktop) and Chrome on Android (tablet), both over HTTPS
- Control: native SDRoxide client (Windows), same server, same radio — **not affected**, used as a working baseline

## Description
Voice transmit over the web client produces no modulated audio on the air, even though every other part of the TX chain works correctly:

- PTT correctly keys the CAT rig (confirmed both by the radio's own TX indicator and by the CAT PTT command reaching it — the transceiver visibly switches to transmit).
- The browser console logs a successful microphone initialization on page load:
  ```
  audio_bridge.js:65  sdroxide audio: context state running
  audio_bridge.js:80  sdroxide audio: playback ready at 48000 Hz
  audio_bridge.js:113 sdroxide audio: mic ready
  ```
- Microphone permission is granted (verified explicitly in Chrome's per-site settings, not just the OS-level permission) on both test machines.
- No JavaScript errors appear in the console at any point, including during PTT press/hold/release.

Despite all of the above, nothing reaches the server: holding PTT and speaking produces **zero** outbound audio frames on the WebSocket, for the full duration of the over. Releasing PTT correctly sends the PTT-off control frame, so the over ends cleanly — it just carried no audio.

Reproduced identically on two unrelated client machines/OSes (Windows 11 desktop, Android tablet), both using Chrome, both over a valid secure (HTTPS) context — which is why the browser's own "audio needs a secure context" gate is not in play here (receive audio, which needs the same secure context, works perfectly on both). The one variable that changes the outcome is the client itself: the **native** SDRoxide client, against the same server and the same radio, transmits microphone audio correctly on the first try. This points at the web client's JS/WASM audio bridge specifically, not at the server, the CAT/Audio backend, browser permissions, or the network path.

## Evidence: HAR capture of the main radio WebSocket during a PTT-held-and-released test
Exported via Chrome DevTools (Network → WS → Messages → Export HAR with content) while pressing PTT, speaking for ~5 seconds, then releasing.

Out of **12,839** total WebSocket messages on the connection, only **10** were client→server (`send`); everything else was server→client (spectrum/audio streaming down). All 10 sends:

```
t=...581.058  8 bytes   EgCSAQAA          (initial handshake-ish)
t=...605.094 32 bytes   EgQJc2ViYXN0aWVuCjAxMjM0NTY3ODk=   (sign-in: "sebastien...")
t=...605.114 52 bytes   EgEtgIABgBA2PAAAAAD9ebvC/XmLwgEAAAAAmJ5pQf59IM8mKGtB
t=...605.114 28 bytes   EgEuAegCUudiUFpB5oBytvHZW0E=
t=...605.432 52 bytes   EgEtgIACgBA2PAAAAAD9ebvC/XmLwgECgt8w65BZQf59IM8IpFxB
t=...605.432 28 bytes   EgEuAQHBb5iyVVpB/z6QZ0HfW0E=
t=...606.183 52 bytes   EgEtgIACgBA2PAAAAABdmrvCXZqLwgECgt8w65BZQf59IM8IpFxB
t=...606.703 52 bytes   EgEtgIACgBA2PAAAAAAc/bvCHP2LwgECgt8w65BZQf59IM8IpFxB
t=...631.263  8 bytes   EgEaAQ==   <- PTT ON  (decodes to 12 01 1a 01)
t=...636.064  8 bytes   EgEaAA==   <- PTT OFF (decodes to 12 01 1a 00, one byte differs)
```
(The six 28–52 byte frames around t≈605 are earlier tuning/panadapter-drag traffic, unrelated to this PTT test — included for completeness.)

The last two frames are ~5 seconds apart, matching the PTT hold duration exactly, and their payloads differ by exactly one byte (`1a01` vs `1a00`) — clearly a PTT-state toggle. **Between these two frames — i.e. for the entire ~5 second over, while speaking directly into the microphone — not a single additional frame was sent.** No audio payload, binary or otherwise, appears anywhere in the capture during the held PTT.

## Impact
Voice (SSB/AM/FM) transmit is completely unusable from the web client — PTT keys the rig, but the operator transmits dead air for the whole over with no indication anything is wrong (no error, no warning, `mic ready` logged as if everything is fine). Anyone relying on the web client as their only access point (e.g. operating remotely without the native client installed, or from a device where installing it isn't practical — a tablet, a borrowed PC) cannot make a voice contact at all, silently.

## Suspected area
`audio_bridge.js` (or the WASM audio glue behind it) appears to successfully obtain the `getUserMedia` stream (hence `mic ready`) but the pipeline that should feed that stream's samples into the outbound WebSocket message stream during an active PTT never actually engages — possibly a wiring gap between the mic `MediaStreamTrack`/`AudioWorklet` node and whatever normally triggers on PTT-down to start pulling frames from it, or a promise/callback that resolves but is never connected to the TX encoder path.

## Reproduction steps
1. Open the web client over HTTPS (`https://<server>/`), sign in.
2. Select a CAT/Audio radio capable of voice TX (SSB tested here).
3. Confirm `mic ready` appears in the console on load.
4. Open DevTools → Network → WS, select the radio's WebSocket connection, watch Messages.
5. Hold PTT, speak for several seconds, release PTT.
6. Observe: PTT-on and PTT-off control frames appear as expected; the radio keys; no audio frames appear between them.

## What we ruled out
- Browser mic permission: granted and confirmed at the per-site level (not just OS-level) on both test devices.
- Secure context: both clients on HTTPS with a valid cert; receive audio (which needs the same `AudioWorklet` secure-context gate) works normally on both, so the page is definitely in a secure context.
- Server-side CAT/Audio configuration: `radio_audio_in`/`radio_audio_out` pinned to the exact ALSA device (confirmed via server logs, `device=... running` with no fallback warnings); CAT link and PTT confirmed working via server logs (`CAT link open`, `rig identified model="TS-2000"`).
- Device/driver/OS specifics: reproduced identically on Windows 11 and Android, ruling out a platform-specific mic driver or permission quirk.
- The CAT rig itself: its own physical microphone (tested directly, not through sdroxide) works perfectly, and the native sdroxide client transmits this same rig's mic audio correctly over the same server/network path.

## Reproduction status
Reliably reproducible, 100% of attempts, on both test machines. Happy to provide the full HAR file, or test a build/patch if useful.

73 and thank you so much

madmedicnl:
Fixes #468.

The browser client's microphone capture never produced a frame. `startMic` fed the worklet (`src.connect(capture)`) but never connected it onward, and a Web Audio node with no path to `AudioContext.destination` is not pulled by the render graph — so `MicCapture.process()` never ran, its `postMessage` never fired, `pullMic()` returned an empty buffer every time, and `micReady`/"mic ready" were logged before any of it could work. PTT keyed the rig and transmitted dead air, with no error and nothing in the console.

This adds the missing `capture.connect(ctx.destination)`. The processor writes no output, so the path carries silence and cannot feed back — the same shape the Web Audio microphone examples use.

**Testing:** the file is JavaScript, so no Rust test covers it. I checked its syntax (node) and the reasoning against #468's HAR (10 client→server frames, none between the PTT-on and PTT-off pair). It needs an off-machine check with a real microphone: the reporter can confirm, or anyone with the web client and a browser on HTTPS.


dividebysandwich:
I cannot reproduce this issue. I ran it in headless Chromium 153 and pullMic() was polled exactly like the WASM client with over 140.000 samples in 3 seconds. I also got the same console output as you.

Edit: Ooooh I should've read the bug report more thoroughly! Got it!

PR #479, opened 2026-09-17
HD Radio: a bench example that decodes a capture

madmedicnl:
A bench harness for the HD Radio decoder, contributed by @dielectric-coder in #466. #466 itself is superseded by the HD Radio work now on `main`, so this is the one piece of it that was not absorbed: the tool that produced the on-air numbers.

```
cargo run --release -p sdroxide-nrsc5 --example hd_capture -- \
    cap.cf32 <capture centre Hz> <channel Hz> [capture rate] [programme]
```

It reads an `sdroxide --record-iq` FM capture at whatever rate the front end ran (2,976,750 S/s — four times the decoder's rate — is convenient), shifts the wanted channel to zero, low-passes and decimates; or a clip already cut to one channel at 744,187.5 S/s, where none of that applies and a `.cs16` extension is read as interleaved 16-bit (half the bytes on disk). Then it drives `HdDemod` directly and prints what the decoder says about itself: lock, per-sideband MER, CBER, carrier offset, PSMI, station name and slogan, the service list, the stereo side channel's RMS, how much audio came out, and the backlog drop count — zero on a healthy stream, and the number that pinned the stereo-pair pacing bug.

**Testing:** `cargo build --release -p sdroxide-nrsc5 --example hd_capture` succeeds here against the current `main`. Running it needs a capture; the decoding path it exercises is the same one the crate's unit tests and the merged decoder cover, so it is a bench aid rather than new decoder surface.


dividebysandwich:
This should already be in, please check if I missed something. Thanks!

madmedicnl:
Closing — this is now redundant. #466 was merged (`dd831560`), which already carried the capture example, and @dividebysandwich followed it with `f3c4712f` ("wait for the decoder thread between blocks rather than overflowing its queue at disk speed, and pick the programme once the station has announced it"). So the example is on `main` and improved; nothing here is needed.

Noted on the two things in the comment on #466: improvements go into new PRs, and anything still being worked on goes up as a draft. The open PRs from the fork are finished changes rather than work in progress, so they are not drafts — but if any of them is better held while something else lands, say which and I will convert it.

dielectric-coder:
Agreed on closing — and confirming the check @dividebysandwich asked for: nothing was missed. `main` carries the example from `1e0621e7`, `c047a194` and `f3c4712f`, and this branch predates the last two, so merging it would have taken the example backwards.

One thing worth recording while the clip was still on the bench, since it is not obvious from the commit message: **`f3c4712f` fixed a real bug, not only the queue overflow it describes.** I ran both versions against the published 6 s CITE-FM clip (`sha256:531ee216…`), same binary otherwise, asking for each of the three programmes in turn:

| programme asked for | `dd831560`, before that commit | `f3c4712f`, `main` today |
| --- | --- | --- |
| HD1 | side channel RMS 0.0401 | 0.0407 |
| HD2 | 0.0401 | 0.0024 |
| HD3 | 0.0401 | 0.0106 |

Before it, the programme argument had no effect — the same output three times, which is HD-1 whatever is asked for. `select_program` ran before the multiplex had announced its services, and the demod ignores a programme it has not been told about, so the call was a no-op. Selecting after the announcement, as it now does, gives three genuinely different programmes. Lock, MER 12.4 / 12.2 dB, CBER 0 and `backlog drops: 0` in every run, on both versions.

That table only covers what I measured today against `main`'s own history; I have not tried to reconcile it with the figures I posted from the fork's branch yesterday, where HD2 already read 0.0024.

PR #485, opened 2026-09-18
Modes: remember the operator's settings per mode

madmedicnl:
AGC, squelch, noise reduction, the auto-notch and the two stereo switches are one set of values for the whole radio today, so switching from a weak-signal data mode to SSB leaves the settings where the last mode wanted them. This makes them per mode: each mode has a set of starting values, the operator's changes while a mode is selected are remembered against that mode, and a **DEFAULTS** chip appears beside the controls to put the mode's own values back.

**The defaults** (`Mode::default_profile`, `crates/sdroxide-types/src/profile.rs`) are deliberately conservative:

- **AGC** slow on the weak-signal digital modes (FT8/FT4/FT2/JS8/WSPR/PSK/RTTY/Olivia/THOR/FSQ/Hell/RF Paint/packet/NAVTEX/WEFAX), stock medium elsewhere.
- **Noise reduction off** everywhere. It carries a make-up gain and an operator's taste, so it is opted into per mode rather than defaulted on — an earlier attempt to default it on moved the audio level on mode change, which the external-audio tests caught.
- **Squelch open**, **auto-notch off** (it cancels constant tones, and in CW/RTTY the signal *is* one), stereo on.

**Behaviour**

- Applied when the mode actually changes and at startup — not when a rig reports the mode it is already in, so a CAT echo does not undo a tweak.
- A change while a mode is selected is stored as that mode's override and trimmed against the defaults, so putting a setting back where it started forgets it.
- Stored per radio in `modeprofiles.json`, flushed on the session tick and on the way out, and only read/written by an engine that remembers its settings.
- Revert: the DEFAULTS chip in the filter/noise row (shown only when something differs, and it names what), and Settings → General → Per-mode settings → RESET EVERY MODE.

`PROTO_VERSION` is 154: `Command::ResetModeDefaults` is appended (`None` = every mode), so no surviving discriminant moves but a v153 peer has no name for it. The profiles travel as receiver state, so nothing else on the wire changes.

Tests: the profile vocabulary (`ModeProfile` overlay/trim, every mode has defaults, unknown keys round-trip) and four engine tests (defaults applied on a mode change, per-mode memory, reset, and surviving a restart). Not tested: the chip and the Settings button in a running GUI beyond a local build, and a remote client driving the reset.


dividebysandwich:
Thanks for the PR!  I noticed that an upgrade wipes user settings. At startup the per-mode values are applied after the saved session is restored. With no overrides saved yet, the mode's defaults win for every setting. So on the first launch after upgrading, every operator's saved AGC, squelch, NR, binaural and RX gain go back to defaults, with no warning. After that, the session's copies of those settings are never used; the comment just above ("the remembered session always wins") now says the opposite. Could you perhaps store the restored session values as the override for the session's mode, or skip the startup step when a session was restored?

madmedicnl:
Fixed, and rebased onto current `main` (the v154 claim collided with the ACARS merge, so per-mode settings are now v155).

**The upgrade wipe.** You were right: the per-mode block ran after the session restore and `effective()` speaks to every field, so with `modeprofiles.json` still empty a first launch after upgrading reset the saved AGC, squelch, NR, binaural and RX gain to the mode's defaults. The profile now applies when a mode is **chosen**, not when the engine starts into a mode the restored session already describes; a first run with no session still gets the mode's defaults.

One thing underneath it: `remember_session.then(|| store.load_session())` returns `Some(Session::default())` even when there is no file, and `Session::default()` passes `is_usable()`, so "restored" and "first run" were the same value. `Store::load_session_if_present` now checks the file is actually there, which is what the startup decision needs — and `load_session` stays as it was so an engine still starts remembering from its first change.

Added `a_restored_session_is_not_reset_by_the_modes_defaults`, which reproduces the upgrade case (a session with a departure, `modeprofiles.json` removed) and fails against the old code. The rest of the mode-profile tests still pass, including the override-survives-a-restart one, which needed no change.

One note for CI: `crates/sdroxide-radio/tests/converter.rs::the_dial_stays_on_air_while_the_hardware_takes_the_offset` fails on plain `upstream/main` here too, so it is not from this branch.

PR #499, opened 2026-09-19
(tr)uSDX: a Kenwood-dialect CAT family for the sound-card setup

madmedicnl:
Adds the DL2MAN/PE1NNZ **(tr)uSDX** as a CAT family.

It emulates a Kenwood TS-480 and answers `ID;` with `020`, so a Kenwood profile seemed right. It isn't: the command subset is thin — dial, mode, PTT, RIT/XIT and VOX, and little else. No S-meter, no SWR, no power control, no VFO B, no split, no keyer: every one of those reads earns a `?;`. So it gets a profile that asks only what this firmware answers, confirmed against a bench radio (the published manual lists a superset of what the parser actually reaches).

This is the **ordinary CAT-rig arrangement**: control over the USB serial port, audio from an external USB sound card wired to the radio's 3.5 mm speaker/mic jack. The radio can also carry its audio *inside* the CAT link as an 8-bit stream with no sound card at all; that needs a different transfer with its own framing and firmware constraints, so it is deliberately not this profile — the stream is switched off at open (`UA0;`) so a session left streaming by another driver cannot pour audio into a `;`-framed parser.

### One hardware fact, asserted rather than configured

Opening the port may reset the radio, and toggling DTR resets it again: the CH340 adapter's DTR is wired to the processor's reset on the common board. DTR is therefore held high for the whole session and never offered as a keying line — a PTT on DTR would reboot the radio on every over. This is the only new driver machinery (`Protocol::holds_dtr_high`).

### Testing

- Unit tests for the frames, the `IF` transmit flag by position, the mode mapping, split-frame replies, and that only the commands the firmware answers are claimed.
- The command set was verified against the real radio on the bench: `FA`, `MD`, `TX;`/`RX;`, `ID020`, `PS`, and the `?;` refusals for `SM`/`RM`/`PC`/`FB`/`RT`/`XT` and the rest.
- `cargo test -p sdroxide-cat -p sdroxide-types` green; fmt clean; clippy shows only a pre-existing warning.

---

This is the sound-card half of a pair. The other PR, #498, adds the same family as a one-cable link that carries the audio inside the CAT stream — take whichever fits.


madmedicnl:
Superseded by #498, which adds the same family and lets the operator choose the audio path — one cable (audio in the CAT stream) or a USB sound card on the 3.5 mm jack. There was no reason to make you pick between them.

Issue #276, opened 2026-09-01
WEFAX - Automatic phasing


fortin-didier-lab:
Hi. Except mistake from my side, it seems with the WEFAX reception mod that the synchronization doesn't work correctely. Indeed, it is mentioned "phasing" once the fax received but when a new fax starts, no more phasing mentioned. The result is the page is shifted.

Beside this small issue, SDROXIDE is absolutely great !!!! Congratulations !!!!

<img width="632" height="102" alt="Image" src="https://github.com/user-attachments/assets/e2135f2c-b277-4a48-8885-70e46d3e67a0" />

<img width="1547" height="1119" alt="Image" src="https://github.com/user-attachments/assets/33214f4e-c32a-415f-a973-8283d12fe075" />

fortin-didier-lab:
Thanks a lot ! Indeed this is solved and so fast ! Waouh ! 
Once more I'm positively surprised by the quality of your this application.

I have some dev proposal which could be very intersting for you . How could we discuss it between us ?

PR #500, opened 2026-09-19
WEFAX: require real evidence before auto start/stop cuts a picture (#496)

### LINKED ISSUE #496: WEFAX: AUTO START and AUTO STOP falsely trigger mid-transmission, breaking maps into fragments
Bug Description:
When receiving WEFAX transmissions (e.g., DWD Pinneberg on 13882.5 kHz) with AUTO START and/or AUTO STOP enabled, the decoder repeatedly interrupts the active reception session and starts a new one. This results in single weather maps being cut into multiple small image fragments (as shown in the attached screenshots).

When receiving completely in manual mode (both AUTO START and AUTO STOP turned off, using manual START / STOP), the image is received cleanly without any line drops or interruptions.

Image

To Reproduce:

Tune to a WEFAX signal (e.g., DWD Pinneberg).

Enable AUTO START and AUTO STOP (or just AUTO START).

Wait for a map transmission to begin.

Observe the receiver window: HF fading (QSB) or high-contrast graphics inside the map body cause the state machine to misdetect start/stop tones, resetting the decoder prematurely.

Expected Behavior:

Once AUTO START detects a valid Start Tone (300 Hz) and begins decoding, it should lock into the active session state and ignore further start tone detections until the session is formally ended.

AUTO STOP should require a continuous, sustained Stop Tone (450 Hz) for a set duration (e.g., 3–5 seconds) before terminating the session to prevent false triggers caused by noise bursts, fading, or image contrast patterns.

Environment / Hardware:

SDR: RSP1A

Software: sdroxide

madmedicnl:
Fixes #496.

## The bug

With AUTO START and/or AUTO STOP on, one chart is cut into fragments; with both off the same transmission decodes cleanly. The DSP and rendering are fine, so the fault is the auto-detection accepting things that are neither tones nor phasing signals.

Two mechanisms:

- **Tone runs are short.** Three windows is 1.5 s. Fine for a real five-second start/stop tone, but also something a fade or a burst of high-contrast picture can hold, and every false accept ends a healthy page.
- **The mid-picture rephase is lenient.** The fix for a missed stop tone (#276) calls a line "phasing" if it is mostly black with a white patch *anywhere*. That is exactly what a dark satellite image, or any chart with one heavy graphic, looks like — and a patch that does not move passes line after line, so four of them cut the picture.

## The fix

- A start signal heard **while a picture is being drawn** needs six windows (3 s) — twice the idle requirement. The real five-second tone still meets it with two seconds to spare.
- The **stop tone** needs seven windows (3.5 s) of a five-second tone.
- The **mid-picture rephase** needs eight lines (4 s at 120 LPM) instead of four, and its shape test is now strict: the line must be almost black, the patch nearly white, and the patch must sit at the **end** of the line, where a phasing pulse belongs — the part a chart cannot fake without a border. Phasing proper keeps the lenient test, where the pulse is expected and a noisy line should still align.

The idle start-tone path and the `#276` behaviour (a new transmission ending the previous page without a stop tone) are preserved.

## Testing

Three tests were written to reproduce the issue first; all three cut the page before the change and none do after:

- `a_dark_chart_with_a_bright_feature_is_not_a_new_transmission`
- `a_brief_stop_tone_burst_does_not_end_the_picture`
- `a_brief_start_tone_burst_does_not_cut_the_page`

The existing `#276` rephase test is unchanged in intent; its fixture now uses a realistic phasing length (a real phasing signal is ~30 s), since the rephase spends eight of those lines recognising the new transmission before the new page phases afresh.

`cargo test -p sdroxide-dsp -p sdroxide-digi` green; fmt clean; no new clippy lints.


dividebysandwich:
Slight problem here, this regresses #276. The strict mid-picture test adds strict && best.1 + 2 * pulse < n. The pulse must sit in the last 10 % of the line buffer. But that buffer is sliced on the old transmission's line clock, so the next transmission's phasing pulse lands at an arbitrary offset. That's precisely why wefax.rs around line 559 tracks rephase_skew between lines rather than checking position: it was built for an arbitrary-but-consistent pulse.

madmedicnl:
You're right, and thank you for catching it. The positional requirement is gone.

The pulse's offset mid-picture is arbitrary because the buffer is cut on the *old* transmission's clock, exactly as you say, so checking for it at the end of the buffer rejected real rephases. What the strict test keeps is shape — the line nearly black, the pulse nearly white and narrow — and `note_phasing_line`'s cross-line consistency is what rejects picture content that moves. I added a #276 test that feeds the next transmission's phasing at an arbitrary 30% offset; it fails if the positional check comes back.

One consequence worth being explicit about: with position gone, a *narrow, clean, static* white stripe on black is indistinguishable from a phasing pulse by the line data alone. So #496's mid-picture protection now rests on the shape test turning away *broad* bright graphics (a chart's cloud mass, a filled border, a heavy graphic) plus the longer eight-line run, not on where the pulse sits. I widened the dark-chart fixture from a 5% to a 25% band to reflect that — the 5% fixture only beat the lenient test by being a perfect phasing pulse, which is the degenerate case I can't separate.

`cargo test -p sdroxide-dsp -p sdroxide-digi` green. If you'd rather the mid-picture rephase lean on the start tone alone and drop the phasing fallback, say the word.

madmedicnl:
Landed upstream — the branch is in `main` now (rebased as `40e76bea`…`4920fcb8`, plus your comment tweak on the shape test). Closing to keep the list tidy; thanks for the review, it made the fix better.

PR #509, opened 2026-09-20
HFDL: an ARINC 635 ground-network decoder (issue #497)

### LINKED ISSUE #497: Add HFDL (High Frequency Data Link) decoder with map tracking
Is your feature request related to a problem? Please describe.
Currently, there is no built-in decoder for HFDL (ARINC 635) transmissions on HF bands.

Describe the solution you'd like:
Add a native HFDL decoder panel to decode aircraft transmissions on HF (3–22 MHz) with real-time map visualization.

Key features:

Automatic decoding of HFDL burst transmissions (M-PSK modulation).

Parsing of flight numbers, aircraft IDs, position reports (lat/lon), and system messages.

Interactive Map view: Display detected aircraft positions, flight paths, and ground stations directly on a built-in map (similar to Dump1090 / ACARS mapping tools).

Decoded message log table with export capabilities.

Additional context:
HFDL is widely used for long-range transoceanic and polar flight tracking where VHF ACARS or 1090 MHz ADS-B are out of reach. Combining HFDL decoding with an interactive aircraft tracking map would make sdroxide an incredible, top-tier tool for HF monitoring!

madmedicnl:
Fixes #497.

## What it is

An HFDL (ARINC 635) decoder — the HF aircraft datalink: ground stations on the shortwave band talking to aircraft over the ocean, carrying squitters, logons, positions, performance/frequency data and ACARS.

The hard part already exists under a permissive licence: **`airframesio/xng` is MIT/Apache-2.0**, and its `xng-mode-hfdl` crate does the demod, FEC and PDU parsing. Its `PROVENANCE.md` records a clean-room implementation from ICAO Annex 10 / ARINC 635 with dumphfdl consulted as facts only, so it is safe to depend on from this GPL-3 project. The dependency tree is small (rustfft, num-complex, chrono, serde, crc; no protobuf). It is a **git dependency** on `xng` at a pinned rev — say the word and I will vendor/port the core into a `sdroxide-hfdl` crate instead, which is the alternative you left to us on the issue.

## How it fits the radio

A `Mode::Hfdl` lane like ADS-B/VDL2/AIS: an engine-owned 24 kHz downconversion of the raw I/Q, no audio, no transmitter, decoded off the `on_rx_iq` tap the other lanes already use. Because HFDL is a published plan of assigned frequencies across 2.8–22 MHz rather than one worldwide channel, the panel chooses the frequency and the dial follows it; the lane is centred on `HfdlSettings::frequency_hz` with the +1440 Hz USB subcarrier handled inside the decoder.

The panel docks under the waterfall like the other lane panels, split into a decode log and an aircraft map: xng already lifts a normalized `details.position {lat, lon, aircraft_id, icao, flight}` out of a performance-data (0xD1) or frequency-data (0xD5) HFNPDU and drops the all-zero not-yet-acquired fix, so the worker parses that into a typed fix the log shows and the map plots (latest fix per identity, retired after 30 minutes of silence).

## Protocol

`PROTO_VERSION` 158 -> 159: `RadioState::hfdl` is appended to `RadioState`'s tail, and `RadioState` crosses whole, so a v158 peer reads the extra bytes as the start of the next field and fails to decode every state — the same break the other appended decoder fields caused, and the same fix. The decode log is engine-side only (`RadioEvent::HfdlStatus`), bridged nowhere.

## Testing

- `cargo test -p sdroxide-hfdl -p sdroxide-dsp -p sdroxide-radio` green, including the position/event parsing and the mode tables.
- The demod chain was validated off-air against a 21 931 kHz capture: it decodes the Riverhead squitter (GS 4), and the ignored test `an_off_air_capture_decodes_through_the_worker` runs it end to end through the worker. The capture itself stays out of the tree (programme material).
- **Not validated on real aircraft traffic here**: the off-air capture carries a squitter and no aircraft position, so the map is unit-tested only. I would rather say that than imply it has been watched working.


dividebysandwich:
Hi, could you please vendor xng in so we get less network dependency?

madmedicnl:
Done. `xng` is now a pinned submodule at `vendor/xng` (rev `096c805`), like `vendor/rade_c` and `vendor/rtl_433`, and `xng-mode-hfdl` is reached by path instead of cargo's git fetcher — so the build takes no network dependency of its own. `Cargo.lock` no longer carries the git source; workspace build, wasm check and the `sdroxide-hfdl` tests are green.

madmedicnl:
A note for review: the lane-rate fix on this branch (the resampling to exactly 24 000 Hz) has now been confirmed on real hardware by an independent tester on the upstream issue.

The decoder only decodes a lane at exactly 24 000 Hz, and our DDC reaches that only when the device sample rate divides evenly into it — 1.536/3.072 Msps did, 2.000 Msps did not (25 000), so an RSP1A at 2.0 Msps got a live waterfall and no decodes at all. With the fix, the same reporter ran **2.000 Msps deliberately** on 21.982 MHz and got **47 decodes and 2 aircraft** (KNE578 at 1980 km, 8 dB). His earlier success on 1.7.3 was his radio happening to be on a rate that divided evenly.

The engine resamples the DDC output onto `HFDL_LANE_RATE_HZ` before the controller sees it, and always builds the controller at 24 000. Pinned by an ignored test against the off-air capture: the same signal at a genuine 25 000 Hz lane decodes 0 events, at 24 000 Hz decodes the Riverhead squitter.

(Full thread on #497.)

madmedicnl:
Just noting for tidiness: this one's work is already in `main` (`bbc47e0a` and the review commits around it — a manual section, rustfmt, doc fixes), so the PR shows OPEN only because the changes landed as commits rather than via the merge button. Safe to close whenever you like.

Two follow-ups to the lane are now offered separately as **#524** — a sample-rate bug and an off-state clarity fix, both reported on #497 after people started using it. Thanks again for the review work on this one; the manual section in particular was the right call.

Issue #511, opened 2026-09-20
Why can't I double-click on the freq and enter it in khz bij hand?


ON5HB:
I would like to be able to double-click the freq-screen and then get a cursor to enter the frequency by hand.
I have not found a simple way of doing this.

dividebysandwich:
Will be in the next release!

PR #512, opened 2026-09-20
Frequency readout: double-click to type a frequency (#511)

madmedicnl:
Addresses #511.

The desktop readout only stepped digits — the type-in editor the compact layout opens with a tap (`show_typed`) was unreachable from it. A **double-click on any digit** now opens that same editor, sharing its MHz prefill, parser and focus-on-open, so there is one editor either way.

The first click of the double has already stepped the digit under the pointer, so the frequency as it was *before* that step is remembered and restored as the editor opens: a double-click that is then cancelled (Escape, or a click elsewhere) leaves the dial where it was rather than one place value away.

The editor takes **MHz** (e.g. `14.074`) — that is what the field has always been labelled and what the compact layout uses. If you would rather it also took kHz, that is one function (`parse_mhz`) and a bare whole number could be read as kHz; say the word and I will fold it in.

Not covered by an automated test: the change is an egui double-click path, and the existing `freq_display` tests (the MHz parser, the prefill round-trip, the digit-count gates) still pass. Happy to add a headless frame test if you want one.


madmedicnl:
Thanks for taking this — I see it went in as `43ce0543`, byte-for-byte the same change as the branch here. Closing the PR since it's in `main` now.

(For anyone reading this later: applied directly rather than merged, which is why the PR stayed open.)

PR #513, opened 2026-09-20
LimeSDR: fold separator spelling in the board allow-list (#508)

### LINKED ISSUE #508: LimeSDR Mini not recognised
trying to get my limesdr mini to work with SDRoxide below is the error report and the output of limeutil --find

### radio session (open / stream / transmit)
     28 ms  LMS_Open                 (first found)                                FAILED: no LimeSDR found — is one plugged in, and does LimeUtil --find see it? (LimeSuite listed 1 non-Lime device(s), which were ignored)

### probe
LimeSuite 23.11.0-gc2d9e877
     28 ms  LMS_GetDeviceList        LimeSDR Mini, media=USB 3, module=FT601, serial=1D424CAEE0153C, index=0 not a Lime board, ignored

LimeSDR support is not yet hardware-verified. If something is wrong, paste this into an issue along with the output of `LimeUtil --find`.

C:\Users\hilux>limeutil --find
  * [LimeSDR Mini, media=USB 3, module=FT601, serial=1D424CAEE0153C, index=0]

the soapy drivers find the lime and it seems to work ok 

thanks in advance John

madmedicnl:
Fixes #508.

LimeSuite reports the Mini with a **space** — `LimeSDR Mini, media=USB 3, module=FT601, serial=…` — while `KNOWN_BOARDS` spelled it with a **hyphen**, `LimeSDR-Mini`. `name_is_known` folded case but not separators, so the board was enumerated and then discarded as *"not a Lime board"*: it never appeared in the picker and could not be opened at all.

The allow-list now compares on one key: case-folded, with space, hyphen and underscore read as the same separator. The spelling genuinely varies — the LimeSDR-USB is hyphenated, the Mini is spaced, and LimeSuite's own table carries both `LimeSDR-Core` and `LimeSDR_Core` — so the list holds one spelling per board. `rx_channels` uses the same key, which also drops its duplicate underscore entry.

It stays an allow-list: the bare Cypress FX3 id an unprogrammed RX-888 presents is still refused.

Not verified against hardware (I have no LimeSDR here) — but this change is only the device-string classification, and the tests cover the spaced Mini, the USB, the v2, the PCIe, LimeNET and Core spellings, and the FX3 rejection. Everything about opening and streaming a LimeSDR is untouched, so the Mini should now reach the code that already exists for it.


madmedicnl:
Superseded — upstream has this as `c0f30ffc` (`LimeDevice::canonical_name`). Closing; thanks.

Issue #510, opened 2026-09-20
redpiyaya puresignal


sa7bxw-commits:
redpitaya 125-14
works fine p1
oc oke
mike oke 

 ps is  visible but not funtion  because repitaya bord has rx1 and rx2 and tx1 and tx2
rx1 nomal for recv,  rx2 recv ps coupler   or via relais extra recv signal
when transmiting must rx2 recieve  de coupler
deskhpsdr p1 i take hermeslite  has no problem and   zeus sdr i use anan-ge2  p1 for ps works perfect

hf output oke
system linux mint

026-09-20T10:20:41.640450Z  INFO sdroxide_hpsdr::net: HPSDR: opening 192.168.1.240, requested RX rate 192000 Hz
2026-09-20T10:20:41.640670Z DEBUG sdroxide_hpsdr::discovery: HPSDR discovery: sending P1 + P2 requests to 1 target(s): [192.168.1.240:1024]
2026-09-20T10:20:41.641473Z DEBUG sdroxide_hpsdr::discovery: HPSDR discovery: reply from 192.168.1.240:1024 (60 bytes): board "Hermes (ANAN-10/10E/100/100B)", Protocol 1, MAC 00:26:32:F0:1B:A4, idle [EF FE 02 00 26 32 F0 1B A4 20 01 00 00 00 00 00]
2026-09-20T10:20:42.471408Z  INFO sdroxide_hpsdr::discovery: HPSDR discovery: 1 device(s) found: 192.168.1.240 (P1, Hermes (ANAN-10/10E/100/100B))
2026-09-20T10:20:42.471511Z  INFO sdroxide_hpsdr::net: HPSDR: 192.168.1.240 answered probe: board "Hermes (ANAN-10/10E/100/100B)", Protocol 1, MAC 00:26:32:F0:1B:A4, idle
2026-09-20T10:20:42.471535Z  INFO sdroxide_hpsdr::net: HPSDR: spectrum INVERTED (I/Q conjugated, receive and transmit alike)
2026-09-20T10:20:42.471547Z  INFO sdroxide_hpsdr::net: HPSDR: driving the open-collector outputs from a custom table (22 band(s) with outputs asserted) — check nothing else (amplifier PTT, antenna relays) is wired to those pins
2026-09-20T10:20:42.471593Z DEBUG sdroxide_hpsdr::net: HPSDR: bound local UDP socket 0.0.0.0:50973
[...]

PR #519, opened 2026-09-20
HPSDR: only warn about the IO board's receive input on a board that has one

### LINKED ISSUE #518: pure signal redpitay
maby fo info when i start servetvia console

2026-09-20T16:03:51.943533Z  INFO sdroxide_hpsdr::net: HPSDR: starting Protocol 1 network thread to 192.168.1.240 (board "Hermes (ANAN-10/10E/100/100B)", RX 192000 Hz, TX 48000 Hz)
2026-09-20T16:03:51.943641Z  INFO sdroxide::hpsdr_source: HPSDR source ready: HPSDR Hermes (ANAN-10/10E/100/100B) @ 192.168.1.240 (0.192 Msps), Protocol 1, center 7072970 Hz
2026-09-20T16:03:51.943666Z  INFO sdroxide::hpsdr_source: HPSDR: PureSignal is on — the receiver is the feedback path, so a coupled sample of the transmitter has to reach it during the over (on a Hermes-Lite 2 that is the IO board's PureSignal input). 32 table steps, feedback decimated by 4. Until the loop locks the transmitter is left exactly as it would have been.
2026-09-20T16:03:51.943674Z  WARN sdroxide::hpsdr_source: HPSDR: PureSignal is on but the IO board's receive input is set to "Radio's own input" — unless you have wired the transmit sample in some other way, the T/R switch takes the coupler away for the length of every over and the loop will never lock
2026-09-20T16:03:51.943753Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: stream starting to 192.168.1.240:1024 at 192000 Hz (speed code 2), EP2 at 48000 Hz; board "Hermes (ANAN-10/10E/100/100B)" has no LNA gain register we can drive; board "Hermes (ANAN-10/10E/100/100B)" has no PA register we can drive; priming registers then sending run command
2026-09-20T16:03:51.946628Z  INFO sdroxide_radio::engine: radio interface connected source=HPSDR Hermes (ANAN-10/10E/100/100B) @ 192.168.1.240 (0.192 Msps)
2026-09-20T16:03:51.947456Z  INFO sdroxide_radio::engine: radio source swapped at runtime source=HPSDR Hermes (ANAN-10/10E/100/100B) @ 192.168.1.240 (0.192 Msps) audio_mode=false
2026-09-20T16:03:52.046434Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: receiver attached
2026-09-20T16:03:52.103090Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: first EP6 datagram received — 1032 bytes, 126 I/Q samples [EF FE 01 06 00 00 00 00]
2026-09-20T16:03:52.103149Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: radio status set 0 — C1 00 C2 21 C3 11 C4 15 (firmware/gateware version bytes)
2026-09-20T16:04:03.519440Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: receiver detached
2026-09-20T16:04:03.519485Z  INFO sdroxide_hpsdr::protocol1: HPSDR P1: shutdown requested; stopping the radio's stream
2026-09-20T16:04:03.521022Z  INFO sdroxide_hpsdr::net: HPSDR: opening 192.168.1.240, requested RX rate 192000 Hz
2026-09-20T16:04:04.352416Z  INFO sdroxide_hpsdr::discovery: HPSDR discovery: 0 device(s) found
2026-09-20T16:04:04.352574Z  WARN sdroxide_hpsdr::net: HPSDR: 192.168.1.240 did not answer the discovery probe; reusing what it reported last time — board "Hermes (ANAN-10/10E/100/100B)", Protocol 1. This is normal when reconnecting while the previous connection is still streaming.
2026-09-20T16:04:04.352607Z  INFO sdroxide_hpsdr::net: HPSDR: spectrum INVERTED (I/Q conjugated, receive and transmit alike)
2026-09-20T16:04:04.352620Z  INFO sdroxide_hpsdr::net: HPSDR: driving the open-collector outputs from a custom table (22 band(s) with outputs asserted) — check nothing else (amplifier PTT, antenna relays) is wired to those pins

madmedicnl:
Fixes #518.

`PureSignal` warned that the IO board's receive input was not set to the PureSignal jack whenever `io_rx_input != IoBoardPureSignal`, without checking that the radio *has* an IO board. That value is only ever sent for a Hermes-Lite 2 — it is applied only when the board is Protocol 1, has an LNA gain register, and this is DDC 0 — so on a Hermes/ANAN (the reporter's board is an ANAN-10) the warning names a setting the radio has no input for and the UI does not offer.

This hoists that condition into an `io_board_rx_input` boolean, gates the warning on it, and reuses it for the `io_rx_input` value so the two cannot drift apart. Nothing changes on an HL2; a non-HL2 board with PureSignal on now gets only the general info line (which already names the HL2 input as the example) instead of a warning it cannot act on.

The underlying report is a Hermes/ANAN, where the feedback sample has to reach the receiver by a path the T/R switch leaves alone (the Rx port) — a wiring question rather than a code one, so I've asked on the issue for the hardware and the wiring before doing any more.

Not tested on HL2 or ANAN hardware here — this is a log gate around an existing condition, and `cargo build --release --bin sdroxide` is clean.


madmedicnl:
Thanks — the intent landed in `main` as `29ed7539`. I notice you reworked it, which is an improvement: rather than my inline `board.has_lna_gain()` test, you added a real `HpsdrBoard::has_io_board()` to the trait and used that, so the "does this radio have an IO board at all" question lives with the board abstraction instead of being inferred from the LNA-gain register. Closing the PR since the concern is fixed in `main`.

(The same probe also came with the #510 wording about which receiver PureSignal reads from, and what to wire on a board with no IO board — nice.)

PR #522, opened 2026-09-20
Settings: grey the per-service auto-upload tick while the master switch is off

madmedicnl:
The auto-upload UI has two levels: the master **"Auto-upload each new QSO"** and the per-service **"Auto-upload each new QSO to X"** tick on each service's tab. They read as independent switches, but only the master gates the path — with it off, a ticked service is never pushed, and the operator is left with a lit target and no uploads.

This disables the per-service tick while the master is off, with the reason on hover. The credentials below stay editable, so a service can still be set up before auto-upload is switched on, and the existing "off above" note now also says the ticks are disabled.

Built and `cargo test -p sdroxide-ui --lib` (609) plus the wasm check are clean; the checkbox has not been exercised by hand here.


madmedicnl:
Thanks for taking this — it's in `main` as `e6ffbe4b`, the same change as the branch here. Closing the PR now that it's applied.

